PR #469, opened 2015-08-12
Good enough Per-Monitor DPI support.

espresso3389:
This does not changes the caption bar/scrollbar sizes so it may not be good for some kind of users but anyway the actual texts are rendered correctly.


espresso3389:
Sorry, there're mistakes...:(

PR #470, opened 2015-08-12
Limited Per-monitor DPI support

espresso3389:
On per-monitor DPI environment, the texts on mintty gets very blurry.
This patch make the texts clear although it still could not resize the window caption and scrollbar.

### DIFF
--- a/src/res.mft
+++ b/src/res.mft
@@ -12,11 +12,6 @@
       />
     </dependentAssembly>
   </dependency>
-  <asmv3:application xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
-    <asmv3:windowsSettings xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">
-      <dpiAware>true</dpiAware>
-    </asmv3:windowsSettings>
-  </asmv3:application>
   <trustInfo xmlns="urn:schemas-microsoft-com:asm.v3">
     <security>
       <requestedPrivileges>
--- a/src/winmain.c
+++ b/src/winmain.c
@@ -80,6 +80,45 @@ load_dwm_funcs(void)
   }
 }
 
+static bool per_monitor_dpi_aware = false;
+
+#define WM_DPICHANGED 0x02E0
+const int Process_System_DPI_Aware = 1;
+const int Process_Per_Monitor_DPI_Aware  = 2;
+const int MDT_Effective_DPI = 0;
+static HRESULT (WINAPI * pGetProcessDpiAwareness)(HANDLE hprocess, int* value) = 0;
+static HRESULT (WINAPI * pSetProcessDpiAwareness)(int value) = 0;
+
+static void
+load_shcore_funcs(void)
+{
+  HMODULE shc = load_sys_library("shcore.dll");
+  if (shc) {
+    pGetProcessDpiAwareness =
+      (void *)GetProcAddress(shc, "GetProcessDpiAwareness");
+    pSetProcessDpiAwareness =
+      (void *)GetProcAddress(shc, "SetProcessDpiAwareness");
+  }
+}
+
+static bool
+set_per_monitor_dpi_aware()
+{
+  if (pSetProcessDpiAwareness && pGetProcessDpiAwareness) {
+    HRESULT hr = pSetProcessDpiAwareness(Process_Per_Monitor_DPI_Aware);
+    // E_ACCESSDENIED:
+    // The DPI awareness is already set, either by calling this API previously
+    // or through the application (.exe) manifest.
+    if (hr != E_ACCESSDENIED && !SUCCEEDED(hr))
+      pSetProcessDpiAwareness(Process_System_DPI_Aware);
+
+    int awareness = 0;
+    return SUCCEEDED(pGetProcessDpiAwareness(NULL, &awareness)) &&
+      awareness == Process_Per_Monitor_DPI_Aware;
+  }
+  return false;
+}
+
 void
 win_set_timer(void (*cb)(void), uint ticks)
 { SetTimer(wnd, (UINT_PTR)cb, ticks, null); }
@@ -865,6 +904,20 @@ win_proc(HWND wnd, UINT message, WPARAM wp, LPARAM lp)
     when WM_INITMENU:
       win_update_menus();
       return 0;
+    when WM_DPICHANGED:
+      if (per_monitor_dpi_aware) {
+        LPRECT r = (LPRECT) lp;
+        SetWindowPos(wnd, 0,
+          r->left, r->top, r->right - r->left, r->bottom - r->top,
+          SWP_NOZORDER | SWP_NOOWNERZORDER | SWP_NOACTIVATE);
+        win_adapt_term_size(false, true);
+
+        char buf[256];
+        sprintf(buf, "SM_CXVSCROLL=%d", GetSystemMetrics(SM_CXVSCROLL));
+        SetWindowTextA(wnd, buf);
+        return 0;
+      }
+      break;
   }
  /*
   * Any messages we don't process completely above are passed through to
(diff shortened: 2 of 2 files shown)


mintty:
I would appreciate if you describe the issue that are trying to address, according to the Contribution hints on https://github.com/mintty/mintty or http://mintty.github.io/, before considering a patch. Right now I only have a vague idea what the actual problem might be. Maybe also some screenshots would help.

espresso3389:
I have in dual monitor environment and one is 2560x1440 (WQHD) 27 inch and the other is 3840x2160 (4K) 27 inch. The first one is in 96dpi and the other is in 144 dpi.

So my screenshot is:
screenshot

But in reality, what I see is like:

realfeeling

The window opened in the 4K display works normally, but if I move the window to the WQHD monitor, the window gets very blurry. (Windows automatically scales the window to 150%).

mintty:
I acknowledge that the window should not get blurred; two more questions though:
* How do other windows behave (e.g. Cygwin console or Windows cmd window)?
* Do you think the window should be scaled down as in your first screenshots (assuming with your patch), or should it rather be kept at roughly the same visual size, provided that it would be scaled to adapt and not be blurred?

espresso3389:
> How do other windows behave (e.g. Cygwin console or Windows cmd window)?

cmd.exe adapts its contents dynamically according to the monitor DPI. (it changes the contents when crossing monitor border)

cmd.exe @ 96-dpi
cmd-96dpi

cmd.exe @ 144-dpi
cmd-144dpi

They looks almost same physical size to me.
The biggest issues on my mintty patch is that it does not controls the titlebar/scrollbar sizes. cmd.exe does everything on its own code; it actually does not use Win32's standard window chrome. All the title bars, scrollbars on its own code. Win32 no longer supports us if we really want to support per-monitor DPI correctly...:(

> Do you think the window should be scaled down as in your first screenshots (assuming with your patch), or should it rather be kept at roughly the same visual size, provided that it would be scaled to adapt and not be blurred?

You can see the cmd.exe's screenshot above. The clarity on the higher DPI is totally different because of the proper font size. I think the default scaling of Windows is totally useless and every app should support per-monitor DPI correctly.

espresso3389:
cmd.exe adapts its contents dynamically according to the monitor DPI. (it changes the contents when crossing monitor border)

cmd.exe @ 96-dpi
cmd-96dpi

cmd.exe @ 144-dpi
cmd-144dpi

They looks almost same physical size to me.
The biggest issues on my mintty patch is that it does not controls the titlebar/scrollbar sizes. cmd.exe does everything on its own code; it actually does not use Win32's standard window chrome. All the title bars, scrollbars on its own code. Win32 no longer supports us if we really want to support per-monitor DPI correctly...:(

Do you think the window should be scaled down as in your first screenshots (assuming with your patch), or should it rather be kept at roughly the same visual size, provided that it would be scaled to adapt and not be blurred?
You can see the cmd.exe's screenshot above. The clarity on the higher DPI is totally different because of the proper font size. I think the default scaling of Windows is totally useless and every app should support per-monitor DPI correctly.

mintty:
I've set up dual-monitor extended desktop now but cannot reproduce the issue.
Maybe because my DPIs do not differ? I don't see an option either to explicitly set a DPI per monitor to simulate the problem.
Would the Scaling option „Maintain Display Scaling“ vs. „Customize Aspect Ratio“ be involved?
Any support is appreciated so I can test the patch.

mintty:
Right now, my test fails already because load_sys_library("shcore.dll") returns null.

Would Graphics Options ▶ Panel Fit ▶ … ▶ Maintain Display Scaling make a difference (cannot be reset here...)?

espresso3389:
I want to confirm that you are using using Windows 8 or later.
Per-monitor DPI is only supported by Windows 8 or later (and shcore.dll is found on them only).

mintty:
No, I wasn't. But I had now the opportunity to check the respective set-up.
I was not able, though, to actually configure different DPI for two monitors. The MS UI for this is braindamaged. Not only is the fact that DPI is affected well-disguised, but the UI looks different than shown on some googled pages, and whether or not I checked the option "Allow me to blabla on different monitors" (whatever), I could not change the "scale..." setting on one monitor without being changed in sync for the other.
Good news is, although I wasn't able to reproduce the effect when moving between monitors, I did encounter the same effect while changing the DPI on one monitor. So I'll integrate your patch, thank you.
Also, by the existing implementation of implicit font zooming, two ways of scaling are already available with your patch; if you hold Shift while scaling (even while reconfiguring from the Windows control panel), the virtual screen size will be maintained by scaling the font.

mintty:
By the way, can you explain why in the resource file res.mft, the DPI-aware flag needs to be *removed* to make this work?

espresso3389:
On winmain.c, you can see the following fragment:

```c
HRESULT hr = pSetProcessDpiAwareness(Process_Per_Monitor_DPI_Aware);
// E_ACCESSDENIED:
// The DPI awareness is already set, either by calling this API previously
// or through the application (.exe) manifest.
if (hr != E_ACCESSDENIED && !SUCCEEDED(hr))
  pSetProcessDpiAwareness(Process_System_DPI_Aware);
```

When if we set Process_Per_Monitor_DPI_Aware using SetProcessDpiAwareness,
it will be failed with E_ACCESSDENIED if the executable has a manifest of Process_System_DPI_Aware.

So if we want to support both Process_Per_Monitor_DPI_Aware and Process_System_DPI_Aware and switch them runtime, we should not set Process_System_DPI_Aware on the manifest file. (This is based on my own research so it may be wrong...)

espresso3389:
[SetProcessDpiAwareness function](https://msdn.microsoft.com/en-us/library/windows/desktop/dn302122%28v=vs.85%29.aspx):

> This API fails if the DPI awareness is specified in the application manifest or if this API has been called previously.

mintty:
Released in 2.2.1.

mintty:
@espresso3389 please:
Do you have an idea what might be causing #492?

mintty:
@espresso3389: This issue falls into a trap built by MS, apparently they don't want to allow applications to get this working; automatic DPI scaling for the title bar and explicit DPI handling for applications contents using WM_DPICHANGED is being prevented. See also https://github.com/mintty/mintty/issues/547#issuecomment-218681712.

mintty:
Released tweak in 2.4.0.

mintty:
Tweaked again in 2.4.1.

mintty:
After various problems with font size (#492, #487), I've experimented a bit and committed a tentative patch to hook DPI change handling into WM_WINDOWPOSCHANGED rather than WM_DPICHANGED.
One reason to look for an alternative was that WM_DPICHANGED is not reliable (it can be disabled from the Windows shortcut), another was a new attempt to also solve the title bar resizing problem (#547) (but that failed). I had first tried to hook into WM_EXITSIZEMOVE/WM_CAPTURECHANGED but that did not work (font was resized but window not).
@espresso3389, please check whether the current repository version is good from your point of view (not making a release now before 3-week travelling). There is a minor glitch still when starting mintty on a specific monitor (`-p@1`) and the moving.

mintty:
Revised DPI handling as described above in 2.5.0 (cygwin test version).

mintty:
Released revised DPI handling in 2.5.1, according to the new Windows 10 DPI handling model.

mintty:
Now supporting Windows DPI handling V2 (#774), also scaling Options and context menu properly.

mintty:
Released 2.9.0.

mintty:
Released 2.9.5 with default HandleDPI=2.

mintty:
See #890 about reintroducing the manifest section `<dpiAware>true</dpiAware>` to support DPI scaling in Windows 7. This has become possible after introducing DPI awareness mode V2.

PR #462, opened 2015-07-14
Support building on 64-bit MSYS2

sschuberth:
This brings @Alexpux' patch from

https://github.com/Alexpux/MSYS2-packages/blob/master/mintty/02-mintty-add-msys-64bit.patch

upstream, with a minor change that prefixes the zip_files path with "../".

Issue #420, opened 2015-06-04
Windows 7 taskbar pinned shortcut compatibility


GoogleCodeExporter:
```
I have two pinned mintty:
"C:\cygwin\bin\mintty.exe -- ssh linux" with icon gentoo.ico, let's say it's 
gentoo-mintty
"C:\cygwin\bin\mintty.exe -" with default icon, let's say it's cygwin-mintty

before starting gentoo-mintty, the two icon of pinned shortcut looks just fine. 
but after starting the gentoo-mintty, the icon becomes default icon instead of 
the custom one (gentoo.ico)
-> suggested enhancement 1: adds support for custom app icon from command line

close all mintty, start the cygwin-mintty, taskbar is shown as
[gentoo-mintty (not started)] | [cygwin-mintty (started)]
but after i press alt+F2 (the "New" window feature of mintty), taskbar looks 
like
[gentoo-mintty (started)] | [cygwin-mintty (started)], this is unexpected
it's expected to be
[gentoo-mintty (not started)] | [cygwin-mintty (started)][cygwin-mintty 
(started)] (the two should group next to each other)

it looks like the "New" window mintty is started with command line
"C:\cygwin\bin\mintty.exe" instead of "C:\cygwin\bin\mintty.exe -", which make 
window cannot recognize what this task should group, so it find the first one 
that match only the path part of command line without arguments

-> suggested enhancement 2: when using alt+F2/New feature, fork new mintty with 
exactly same command line argument as current one (don't forget to load correct 
custom icon btw)
```

Original issue reported on code.google.com by `xue...@gmail.com` on 15 Feb 2014 at 4:23

dscho:
> suggested enhancement 1: adds support for custom app icon from command line

Is that not addressed by [the `-icon` command line option](https://github.com/mintty/mintty/blob/271b95218844eac5ba7b85fee3b0124864d099e5/docs/mintty.1#L77-L81)?

mintty:
Issue 1: As dscho noted, the ```-i``` or ```--icon``` option would achieve the desired effect; however, it would be more intuitive and desirable if a dedicated icon configured for the shortcut being pinned would also persist on the taskbar. This is not the case, probably because mintty sets its own icon with the Windows API. A patch that handles this nicely (while keeping the default behaviour) would be appreciated.

mintty:
Issue 2: I cannot reproduce this one. If I pin two different invocations of mintty with two different icons and start a few instances, behaviour is as expected here.

[...]

PR #471, opened 2015-08-16
Better support for pinning to the task bar

dscho:
Windows 7 allows users to pin running applications to the task bar. However, when launching `mintty` from another process, clicking the pinned icon will call `mintty` instead of the executable of the calling process.

To help with this issue, this Pull Request adds support for two new options:

- `RelaunchDisplayName`: override the name shown in the task bar
- `RelaunchCommand` override the command-line to call from the icon pinned to the task bar

Note that Windows will simply ignore the relaunch command setting unless both the display name and the application ID are configured, too.

Please note also that the way `mintty` configures the application ID currently (`SetCurrentProcessExplicitAppUserModelID`) did not work correctly in this developer's setup (Windows Server 2012 R2): when configuring the application ID of the calling process and of `mintty` so they would match, the pinned icon would *still* launch `mintty` directly. This was only fixed by setting the application ID *explicitly* for the `mintty` terminal window (this is what the first commit is all about).

Background: [Git for Windows 2.x](https://git-for-windows.github.io/) is slated to ship with `mintty` as its default terminal emulator for "Git Bash", launched as `git-bash.exe`, and we want `git-bash.exe` to be called when a pinned *Git Bash* icon is clicked by the user.

### DIFF
--- a/docs/mintty.1
+++ b/docs/mintty.1
@@ -990,6 +990,20 @@ be done using the \fBwin7appid\fP utility available from
 \fIhttp://win7appid.googlecode.com\fP.
 
 .TP
+\fBRelaunch Command\fP (RelaunchCommand=)
+Windows 7 and above allow pinning applications to the taskbar. With this
+property, the command line to be called can be customized. Note that the
+Relaunch Display Name and the Application ID property need to be set,
+too, or the setting will be ignored by Windows.
+
+.TP
+\fBRelaunch Display Name\fP (RelaunchDisplayName=)
+Windows 7 and above allow pinning applications to the taskbar. With this
+property, the label of the taskbar menu can be customized. Note that the
+Application ID property needs to be set, too, or the setting will be ignored
+by Windows.
+
+.TP
 \fBWord selection characters\fP (WordChars=)
 By default, this string setting is empty, in which case double-click word
 selection uses the default algorithm that is geared towards picking out
--- a/src/config.c
+++ b/src/config.c
@@ -87,6 +87,8 @@ const config default_cfg = {
   .daemonize = true,
   // "Hidden"
   .app_id = "",
+  .relaunch_command = "",
+  .relaunch_display_name = "",
   .col_spacing = 0,
   .row_spacing = 0,
   .word_chars = "",
@@ -212,6 +214,8 @@ options[] = {
   // "Hidden"
   {"AppID", OPT_STRING, offcfg(app_id)},
   {"ColSpacing", OPT_INT, offcfg(col_spacing)},
+  {"RelaunchCommand", OPT_STRING, offcfg(relaunch_command)},
+  {"RelaunchDisplayName", OPT_STRING, offcfg(relaunch_display_name)},
   {"RowSpacing", OPT_INT, offcfg(row_spacing)},
   {"WordChars", OPT_STRING, offcfg(word_chars)},
   {"WordCharsExcl", OPT_STRING, offcfg(word_chars_excl)},
--- a/src/config.h
+++ b/src/config.h
@@ -107,7 +107,7 @@ typedef struct {
   int x, y;
   bool daemonize;
   // "Hidden"
-  string app_id;
+  string app_id, relaunch_command, relaunch_display_name;
   int col_spacing, row_spacing;
   string word_chars;
   string word_chars_excl;
--- a/src/std.h
+++ b/src/std.h
@@ -41,7 +41,7 @@ int vasprintf(char **, const char *, va_list);
 
 char *asform(const char *fmt, ...);
 
-#define WINVER 0x500  // Windows 2000
+#define WINVER 0x601  // Windows 7
 #define _WIN32_WINNT WINVER
 #define _WIN32_IE WINVER
 
--- a/src/winmain.c
+++ b/src/winmain.c
@@ -47,6 +47,10 @@ typedef struct {
   int cyBottomHeight;
 } MARGINS;
 
+#else
+
+#include <uxtheme.h>
+
 #endif
 
 static HRESULT (WINAPI * pDwmIsCompositionEnabled)(BOOL *) = 0;
--- a/src/winmain.c
+++ b/src/winmain.c
@@ -1268,8 +1268,8 @@ main(int argc, char *argv[])
   }
 
 #if WINVER >= 0x601
-  // Set the app ID explicitly
-  if (*cfg.app_id) {
+  // Set the app ID explicitly, as well as the relaunch command and display name
+  if (*cfg.app_id || *cfg.relaunch_command || *cfg.relaunch_display_name) {
     HMODULE shell = load_sys_library("shell32.dll");
     HRESULT (WINAPI *pGetPropertyStore)(HWND hwnd, REFIID riid, void **ppv) =
       (void *)GetProcAddress(shell, "SHGetPropertyStoreForWindow");
@@ -1290,6 +1290,26 @@ main(int argc, char *argv[])
             var.vt = VT_LPWSTR;
             pps->lpVtbl->SetValue(pps,
                 &PKEY_AppUserModel_ID, &var);
+          }
+        }
+        if (*cfg.relaunch_command &&
+            (size = cs_mbstowcs(0, cfg.relaunch_command, 0) + 1)) {
+          var.pwszVal = malloc(size * sizeof(wchar));
+          if (var.pwszVal) {
+            cs_mbstowcs(var.pwszVal, cfg.relaunch_command, size);
+            var.vt = VT_LPWSTR;
+            pps->lpVtbl->SetValue(pps,
+                &PKEY_AppUserModel_RelaunchCommand, &var);
+          }
+        }
+        if (*cfg.relaunch_display_name &&
+            (size = cs_mbstowcs(0, cfg.relaunch_display_name, 0) + 1)) {
+          var.pwszVal = malloc(size * sizeof(wchar));
+          if (var.pwszVal) {
+            cs_mbstowcs(var.pwszVal, cfg.relaunch_display_name, size);
+            var.vt = VT_LPWSTR;
+            pps->lpVtbl->SetValue(pps,
+                &PKEY_AppUserModel_RelaunchDisplayNameResource, &var);
           }
         }
         pps->lpVtbl->Commit(pps);
--- a/src/winmain.c
+++ b/src/winmain.c
@@ -17,6 +17,10 @@
 #include <shellapi.h>
 
 #include <sys/cygwin.h>
+#if WINVER >= 0x601
+#include <propsys.h>
+#include <propkey.h>
+#endif
 
 HINSTANCE inst;
 HWND wnd;
@@ -1262,6 +1266,38 @@ main(int argc, char *argv[])
     SetWindowPos(wnd, null, 0, 0, 0, 0,
                  SWP_NOMOVE | SWP_NOSIZE | SWP_NOZORDER | SWP_FRAMECHANGED);
   }
+
+#if WINVER >= 0x601
+  // Set the app ID explicitly
+  if (*cfg.app_id) {
+    HMODULE shell = load_sys_library("shell32.dll");
+    HRESULT (WINAPI *pGetPropertyStore)(HWND hwnd, REFIID riid, void **ppv) =
+      (void *)GetProcAddress(shell, "SHGetPropertyStoreForWindow");
+
+    if (pGetPropertyStore) {
+      size_t size;
+      IPropertyStore *pps;
+      HRESULT hr;
+      PROPVARIANT var;
+
+      hr = pGetPropertyStore(wnd, &IID_IPropertyStore, (void **) &pps);
+      if (SUCCEEDED(hr)) {
+        if (*cfg.app_id &&
+            (size = cs_mbstowcs(0, cfg.app_id, 0) + 1)) {
+          var.pwszVal = malloc(size * sizeof(wchar));
+          if (var.pwszVal) {
+            cs_mbstowcs(var.pwszVal, cfg.app_id, size);
+            var.vt = VT_LPWSTR;
+            pps->lpVtbl->SetValue(pps,
+                &PKEY_AppUserModel_ID, &var);
+          }
+        }
+        pps->lpVtbl->Commit(pps);
+        pps->lpVtbl->Release(pps);
+      }
+    }
+  }
+#endif
 
   // The input method context.
   imc = ImmGetContext(wnd);


mintty:
As I do not use the pinning feature myself and have no time to check the details right now (holidays, other patches...), I have only a vague idea of the issue. I acknowledge that your description of behaviour changes sounds reasonable; one question though:
Could the issue not be resolved without introducing new options?

dscho:
> Could the issue not be resolved without introducing new options?

I am afraid not.

The use case I have is a `git-bash.exe` which essentially sets some environment variables that need to be set before MSys2 is allowed to start (and we use `mintty` as an MSys2 program), then calls `mintty.exe`.

Note that `git-bash.exe` does *not* open any window. That is crucial, I think, because when I let it open a message box, the current way to set the application ID did half the job: the display name (i.e. the name displayed in the context menu of the task bar icon) had changed from `Terminal` to `GitForWindows`. This required a windows to be opened, though, pretty ugly. I might get away with opening a hidden window, but that still would be ugly.

However, the bigger issue is that even then a pinned task bar icon would launch `mintty.exe` *directly*, i.e. without going through `git-bash.exe` and naturally skipping my carefully crafted initialization. Indeed, calling `mintty.exe` that way even skips the entire `/etc/profile` because Bash is called without the `--login` option.

The only way to fix that is by setting the `RelaunchCommand` property. Which requires the application ID to be set *explicitly* (i.e. setting it via `SetCurrentProcessExplicitAppUserModelID()` is *not* good enough, and the `RelaunchDisplayNameResource` property as well) otherwise the `RelaunchCommand` property is simply ignored.

So I am afraid we need to set those properties.

Back to your question whether we could get away without introducing new options: Seeing as the application ID cannot contain spaces, we *may* get away by overloading the `AppID` setting to mean "if it contains a space, everything after the space is a relaunch command". But that is pretty ugly, unintuitive and still does not allow us to specify the display name (which can contain spaces, just like the relaunch command).

Therefore I am pretty convinced that I provided the minimally invasive solution.

My only concern with this PR is that you might want to allow developers to compile `mintty` even with Windows 2000 headers. In other words, the `#define WINVER 0x601` might cause trouble. It should not, of course, because `mintty` is compiled pretty much only in the Cygwin and MSys2 context, correct? Both contexts provide up-to-date headers.

dscho:
@mintty anything else I can do to get this PR merged?

mintty:
As I said, I need some time to check the issue. Also one other major patch is in the queue before. Sorry, but you can just have some patience, please.

dscho:
No worries. I was just concerned that you were waiting for something from my side, is all.

mintty:
I have now installed Git for Windows but couldn't reproduce the issue, probably because it ships with a patched mintty already.
So I tried replacing its mintty with my own but had to install MSYS2 first to compile a compatible version. However, gcc and make are not available in the standard MSYS2 installation and how would I know that MSYS2 uses 'pacman'?...
OK, obstacles passed, copied new mintty (without this patch) into Git's /bin and /usr/bin.
Starting from pinned icon and verified version with Options-About.
Yet I do not see what's wrong. I added an echo to Git's /etc/profile and that's being output.
Any further hints on what you actually need to get fixed or how it could be analysed?

dscho:
> I have now installed Git for Windows but couldn't reproduce the issue, probably because it ships with a patched mintty already.

Correct.

> So I tried replacing its mintty with my own but had to install MSYS2 first to compile a compatible version. However, gcc and make are not available in the standard MSYS2 installation and how would I know that MSYS2 uses 'pacman'?...

Yes, the MSys2 documentation mentions somewhere that you need to run `pacman -S base-devel binutils` once before you can start building MSys2 packages.

> Yet I do not see what's wrong.

Since you had it already pinned once, there is now a `.lnk` file in `$HOME/AppData/Roaming/Internet Explorer/Quick Launch/User Pinned/ImplicitAppShortcuts/<some-hex-string>/Git Bash.lnk`.

If you unpin `Git Bash` and then remove that `.lnk` file and the re-pin `Git Bash`, you will see what was described in https://github.com/git-for-windows/git/issues/263: clicking the shortcut will *not* launch a full Git Bash, but instead `mintty.exe` with a plain Bash that was called without the `--login` flag, i.e. it does not know anything about Git, it has no colorful prompt, and it does not have the `MSYSTEM` (and other) environment variables set correctly.

mintty:
I can still not produce any unexpected behaviour. I unpin, remove that file (why would any real user actually do that?), repin, close the open mintty, start a new pinned instance, everything OK, same environment (checking the profile echo, calling env and diff).
I browsed your issue 263 quickly and had the idea that manipulation of needed properties could be done in the wrapper already, before I read this had already been proposed by someone and tried by you – but, if I read correctly, only for the AppID. What if the wrapper sets *all* required properties, including Relaunch and whatever?

dscho:
The problem only occurs if a separate `.exe` calls `mintty.exe` with additional parameters, and if that other `.exe` *does not open a window*. That is the rub, that's why the `AppID` is ignored. And that is why I cannot even set the other window properties: I do not have a window.

mintty:
I could reproduce the effect finally. Previously I had "pinned from a desktop shortcut" which obviously makes a difference. Which makes me again think of other solutions, for example:
What if you create a proper shortcut (like the one I had on the desktop and with which it works) and place that into the Startmenu folder?
Or, even more straightforward: Why don't you put all the required initialization into Git's /etc/profile, or a dedicated profile referred in the shortcut, or something like that? I feel a little bit alienated by the concept of a .exe wrapper just for the sake of setting a few environment things, honestly.
Sorry for the hassle but I don't feel comfortable with adding exotic options unless they were *really* and *undeniably* needed, following Andy's policy of keeping the set of options concise.

mintty:
On further experimenting, pinning is actually also a problem with such an .exe wrapper.
If I simply pin an invocation (whether from a shortcut or from the command line), activating it will skip the profile as in your case.
Is there any additional property that can be set to also maintain invocation parameters?
If your patch can be extended to a more general solution, covering this case, I would be more inclined to integrate it :)

dscho:
AFAICT the `RelaunchCommand` can contain command line options... I have not tested that specifically, though.

mintty:
Fixed previous comment: On further experimenting, pinning is actually also a problem with**out** such an .exe wrapper. The deal offered is: if a patch solves the generic problem, it's more likely to be accepted - win-win for git and mintty :)
I also notice that once this obscure Relaunch property has been applied, Windows seems to remember it somewhere (probably the registry) - nothing pinned anymore, not using a patched version anymore, yet the RelaunchDisplayName is still imprinted on new mintty instances and their icons. Which leaves me with the uncomfortable feeling that this is really Windows black magic (and poorly documented if at all) which I wouldn't like to use without really understanding the implications.

dscho:
> once this obscure Relaunch property has been applied, Windows seems to remember it somewhere (probably the registry)

In my tests, it was both necessary and sufficient to remove that `.lnk` file inside the `$HOME/AppData/Roaming/Internet Explorer/Quick Launch/User Pinned/ImplicitAppShortcuts/<some-hex-string>/` directory to make Windows forget about that property.

> if a patch solves the generic problem, it's more likely to be accepted - win-win for git and mintty :)

I intended this patch to be exactly that: I would not have bothered to submit it if I had the impression that it benefits *only* Git.

Please note, though, that you *still* have to remove the `.lnk` file -- both for Git and in the generic mintty case -- if you want to provide a new Relaunch property.

simnalamburt:
@mintty, if you're going to accept this PR please just merge it and apply your demanding change on the top of the PR, instead of rewriting these commits.

There were several pull requests to this project but those PRs were not merge but [copied](https://github.com/mintty/mintty/commit/9dd74985d93ab167af7beb2f6b3287579d30cf7b). So GitHub still [shows only one contributor](https://github.com/mintty/mintty/graphs/contributors), and this will strongly discourage developers who contributed to this project. You may consider this as very small problem but this is very fatal to OSS development.

mintty:
Thanks for the hint, but as you can see in the commit you quoted, I did include the contributing author explicitly (commit --author=...) as I did on some other occasions of substantial contributions; also the statistics you quote is subtitled "Contributions to master, **excluding merge commits**" which means accepted pull requests would not show up here anyway – maybe a weakness of github that authored commits are not considered there.

There are other – and in my opinion more desirable – ways of honoring contributions, authors are mentioned in the changelog and I also intend to maintain the "Credits" wiki page.
People prefer different paradigms of project handling, and I tried to elaborate on my opinion about merges and branches in #462#issuecomment-12159107.
Often a patch needs to be tweaked or updated before it can be integrated, in which case merging isn't simply possible anyway.

dscho:
> Often a patch needs to be tweaked or updated before it can be integrated, in which case merging isn't simply possible anyway.

I am more than willing to tweak or update if you tell me concretely what you want from me.

mintty:
@Johannes: The remark you responded to was meant more general, discussing the merge proposal.

About this issue, my further experiments give me some more insight in this weird Windows mechanism.
What bothers me is the hidden persistence and its resilience (you talk about deleting a file in some cryptic nested directory which is really insane); there must be some way to reset such a property, I thought, but several attempts (setting empty or null values...) failed.
If an option "AppName" (preferred to "RelaunchName", what do you think?) gets introduced, there should be a straightforward way like -o AppName="" to reset its implications. At least that's what I would expect, and likely other users too, whose complaints and trouble reports I'd prefer to avoid.

Other details are minor, and I can handle them myself unless you also wish a "direct merge" as has been suggested.

dscho:
> there must be some way to reset such a property, I thought, but several attempts (setting empty or null values...) failed.

Yep, I thought the same. If I would have found a more appropriate way to reset this, I would have told you.

> Other details are minor, and I can handle them myself unless you also wish a "direct merge" as has been suggested.

Yes, I would prefer my changes to be merged. If I have to work a little harder to that end, I am prepared to do that.

mintty:
Before proceeding with the relaunch options, 
have you checked my suggestions to achieve the desired effect with a dedicated rc-file?

How do you actually also get the icon persisted for the pinned shortcut?
This failed in my attempts to apply the feature and I would like the solution to work also for the icon if manually invoked.
Maybe there is an additional icon-related property value that should be set in the relaunch setup code?
Also, if you have an idea to alternatively use the icon from an invoking shortcut, I would appreciate that very much.
See #420  "Issue 1".

Discussing the proposal:
I think I'd prefer to call the options AppName and AppCmd (unless you convince me that's not a good idea) and add another pure command-line option (i.e., not configurable in .minttyrc) --setup-pinned as a prerequisite just to make sure this persistence setting is not applied indiscriminately.
(A useful alternative might be to install the pinned shortcut right away, see
http://stackoverflow.com/questions/1968137/windows-7-taskbar-pin-or-unpin-program-links/25812667#25812667)

johannes:
(Please don't write @johannes if you want to respond to @dscho as this will subscribe me and I can't give any useful input. Thanks.)

mintty:
Oops, sorry. Updated that comment and will remove yours and this one.

mintty:
@dscho: I plan a release soon and the issue is not forgotten, but I do not intend to include this patch for now because I’m still not confident with the matter and the suggested solution as there are two major open issues:

Concerning icon setup, I have checked the patch again, getting the result that the patch actually behaves even worse than the default behaviour.
There are 3 icons: shortcut icon (shortcut Properties – Change Icon…), command line icon (-i ...), mintty default icon.
If I make a shortcut with both shortcut icon and -i icon, and pin it, the pinned icon will be the -i icon as long as an instance is running, and flips back to the shortcut icon after it’s terminated.
With git-patched mintty, the pinned shortcut also shows the -i icon while an instance is running, but flips back to the mintty default icon after termination.

The other issue is whether we really need to particularly support the wrapper scenario.
My question was if not a shortcut command like
```...\mintty ... -e .../bash --rcfile .../.gitbashrc```
could achieve the desired setup. In that case, pinning would implicitly maintain the invocation parameters.
If that works, I think it’s a solution to be preferred because it does not involve this extremely weird and insane Windows magic.

Having said this, I have added a function configure_properties() where you could insert your (rebased) patch if needed, and a new option --properties to enable its invocation, as I wish to make this very explicit.
(Also I would prefer any relaunch options to be local to winmain.c, to make them command-line only, so they cannot be configured in .minttyrc; I am willing to work out the details should the patch later be integrated.)

dscho:
> The other issue is whether we really need to particularly support the wrapper scenario.

Well, I do.

> My question was if not a shortcut command like
> ...\mintty ... -e .../bash --rcfile .../.gitbashrc
> could achieve the desired setup.

No, because of the absolute path. That flies in the face of a portable `git-bash.exe`.

> Having said this, I have added a function configure_properties() where you could insert your (rebased) patch if needed, and a new option --properties to enable its invocation, as I wish to make this very explicit.
> (Also I would prefer any relaunch options to be local to winmain.c, to make them command-line only, so they cannot be configured in .minttyrc; I am willing to work out the details should the patch later be integrated.)

I will have a look soon. By now I have forgotten pretty much all of the context and will require a bit of reading before I can do all that.

mintty:
The absolute path of .gitbashrc, you mean? That should be a minor problem. Basically you have a dedicated small cygwin (or msys) environment, with "/" referring to its implicit root when started. So /.gitbashrc or /conf/gitbashrc should work. When you create the Git Bash shortcut, there are multiple ways to tweak it to your needs (e.g. you're already using %HOMEDRIVE%%HOMEPATH%).

dscho:
> The absolute path of .gitbashrc, you mean?

No. You suggested creating a shortcut file instead of a wrapper. And I said that the `Target:` in that shortcut file is *absolute*. Meaning: no more portable Git. If I heed your advice. You cannot have both: *portable* **and** *shortcut*.

And yes, I am a little frustrated because I put in quite a lot of work and it looks like I do not get my problem solved.

> Basically you have a dedicated small cygwin (or msys) environment, with "/" referring to its implicit root when started.

And that is different from what I am doing already how?

> When you create the Git Bash shortcut,

I have no control over that after the user unpacked the portable Git. Thankyouverymuch.

> you're already using %HOMEDRIVE%%HOMEPATH%

OMG. This %HOMEDRIVE%%HOMEPATH% has caused us so much trouble. Hence my code to force %HOME% to something sane in the wrapper. Yep, the same wrapper you want me to abandon.

You leave me in quite tough a spot here.

dscho:
I guess the only resolution of this problem is to abandon the idea of reconciling my changes with mintty? I can easily fork off a `git-bash` from `mintty`. Of course I would prefer not to do that, but it seems that Git for Windows' needs cannot be addressed by mintty. Sad.

mintty:
I am willing to find a common solution once all problems have been sorted out; forking is in general the worst solution.

I had in fact tested with the Git Bash shortcut and not considered you would like a command-line (or GUI) invocation to work alike. I don't know whether a shell script could be established as a wrapper (or would popup a console window...) but in any case, I see and accept you prefer an .exe wrapper for direct invocation.

So now there is only one open issue, the icon problem.
If I start git-bash.exe from the command line, the Git icon will persist and be maintained in the pinned shortcut. If I try to achieve the same running mintty from the command-line and using your patch, it fails (maybe I broke it for testing). If we can solve this issue, I will integrate the patch.

mintty:
Actually, sorry to bother you again with this idea, but my proposal to place your setup in an rc file does not require an additional explicit --rcfile parameter. You only need to place setup into /etc/profile as I had previously done for testing and mentioned in an earlier comment. This will work well with portable Git. Well, provided /etc/profile gets loaded by a login shell; I would offer an option to enforce that which can be placed in /etc/minttyrc (also within the portable installation).

dscho:
> I don't know whether a shell script could be established as a wrapper

I guess you did not mean "shell" because that would be a catch-22.

You probably meant a `.cmd` or a `.vbs`. But those cannot have nice icons. I want a better user experience. And somehow, you know, I found it. `git-bash.exe` can solve those problems. Yes, I had to patch `mintty` to be able to pin to the task bar. But now I have a new problem, don't I.

> If I start git-bash.exe from the command line, the Git icon will persist and be maintained in the pinned shortcut. If I try to achieve the same running mintty from the command-line and using your patch, it fails (maybe I broke it for testing). If we can solve this issue, I will integrate the patch.

I do not quite understand how you run `mintty` from the command-line using my patch. Could you maybe paste an explicit command I can paste into my `cmd.exe` to verify?

> Actually, sorry to bother you again with this idea, but my proposal to place your setup in an rc file does not require an additional explicit --rcfile parameter.

How would that solve my problem? I want a `git-bash.exe` that is double-clickable and that sets up the environment variables and then calls mintty with the appropriate parameters. I do *not* want to see `Terminal` in the task bar. I do *not* want `mintty.exe` to run the Git Bash.

dscho:
Anyway, if you give me some concrete steps to test, with concrete expectations that are unmet, I can try to do something about it. Once I find time again, that is. I really had hoped to resolve this issue when my memory was much fresher on the issues.

mintty:
To avoid confusion by my local adaptation (with static test data), I reapplied your patch to mintty 2.0.1, then started:
```./mintty -o AppID=Test.Mintty -o RelaunchCommand='C:\cygwin\bin\mintty.exe' -o RelaunchDisplayName="Test Mintty" -i /cygdrive/c/Windows/System32/calc.exe /usr/bin/bash --login -i```
The chosen icon appears on title bar and task bar. If I pin it on the taskbar and then close the terminal, the icon flips back.

Well then, if I replace `C:\cygwin...` with `C:\...\Git\git-bash.exe`, the Git icon is used instead. So the pinned icon while no application is running depends on the RelaunchCommand but while it's running it's the actual application icon. This is even more insane as anything I learned before about this issue. But it may also bring us closer to a solution.
It would be nicer to have control over the icon being used. Do you see any chance for that? Any additional property in this ‘AppUserModel’ that could be set? Where is this stuff actually documented? I couldn't find anything about it.

dscho:
Unfortunately I only have hyperthreads available to take care of this issue, but from what you are saying, it looks as if the problem was solved by reusing the `-i` option to set the [`RelaunchIconResource` property](https://msdn.microsoft.com/en-us/library/windows/desktop/dd391573%28v=vs.85%29.aspx), too?

mintty:
I have uploaded a new tweak. Please check; options are --store-taskbar-properties -o AppName=... -o AppLaunchCmd=...
If that's a solution, I shall upload another back-stripped version to insert you as author, or you may provide a rebased patch if preferred.
Documentation pending.

dscho:
@mintty I will have a look tomorrow morning, UTC+0200 time zone.

elieux:
+1 These options would save me a lot of hassle and issues with my [MSYS2 launcher](https://github.com/elieux/msys2-launcher). I'm not sure what's the status of this feature request. If there's anything that needs to be done to get this in the next release, let me know.

elieux:
There's a minor issue with the `AppUserModel_RelaunchIconResource`. It's set to the Cygwin path to the icon (e.g. `/msys.ico`), but it should be set to the Windows path. (I'm testing with current [master](https://github.com/mintty/mintty/commit/f9db71ff256fa7e54a29373b02b9c958a7f72750).)

elieux:
Sorry for the spam, but does anyone have an idea what determines the `Start in` field (i.e. working directory) of the pinned shortcut?

mintty:
> I'm not sure what's the status of this feature request.

The plan is to include the feature in the next release (soon)...

>  but it should be set to the Windows path.

with this fixed, of course; thanks for the hint. And I'll check whether this also solves issue 1 of #420 (setting taskbar icon while unpinned/when not storing properties) which would be great.

> Sorry for the spam, 

This isn't spam and your comments are welcome, especially in this obscure area of Windows magic.

> but does anyone have an idea what determines the Start in field (i.e. working directory) of the pinned shortcut?

No; does it work as you desire if you set a working directory in a shortcut and pin that? Would it make sense if mintty provides a workaround option to set an initial directory? This would, in contrast to any initial "cd" injected to the shell, also affect Alt+F2.

elieux:
>> but does anyone have an idea what determines the Start in field (i.e. working directory) of the pinned shortcut?
> No; does it work as you desire if you set a working directory in a shortcut and pin that?

Create shortcut to a random .exe -> Set `Start in` -> Pin to Taskbar -> Check `Start in` on pinned shortcut -> works.

> Would it make sense if mintty provides a workaround option to set an initial directory? This would, in contrast to any initial "cd" injected to the shell, also affect Alt+F2.

I guess so (for reference/example, ConsoleZ can do this). I don't see any immediate use for this in Cygwin/MSYS2, as their shells always start in the home directory, but maybe @dscho can weigh in from the side of GfW.

mintty:
>>> but does anyone have an idea what determines the Start in field (i.e. working directory) of the pinned shortcut?

>> No; does it work as you desire if you set a working directory in a shortcut and pin that?

> Create shortcut to a random .exe -> Set Start in -> Pin to Taskbar -> Check Start in on pinned shortcut -> works.

I cannot confirm that, it does not work here.
(If it did, my second question would be: can I likewise make use of the icon configured in a shortcut?)

>> Would it make sense if mintty provides a workaround option to set an initial directory? This would, in contrast to any initial "cd" injected to the shell, also affect Alt+F2.

> I guess so (for reference/example, ConsoleZ can do this). I don't see any immediate use for this in Cygwin/MSYS2, as their shells always start in the home directory, 

Not necessarily. It's the shell that decides where to start, as a result of its profile processing.

mintty:
Your earlier comment:
> icon... should be set to the Windows path.

I'm adding a hint to the manual page (rather than converting) and a complete example.

elieux:
> I cannot confirm that, it does not work here.

Are you on Windows 10? Maybe it behaves differently. My last report was from Windows 6.x.

> can I likewise make use of the icon configured in a shortcut?

That should definitely work.

> Not necessarily. It's the shell that decides where to start, as a result of its profile processing.

Yeah, and the Cygwin/MSYS2 shell always decide to start in the home directory by default. Or did I misunderstand?

>> icon... should be set to the Windows path.
> I'm adding a hint to the manual page (rather than converting) and a complete example.

That doesn't sound too good. I find it nice that we can specify icons like `/msys2.ico`. Will Windows-style relative paths work at least?

mintty:
("Start in" folder not cloned into pinned icon)
>> I cannot confirm that, it does not work here.

> Are you on Windows 10? Maybe it behaves differently. My last report was from Windows 6.x.

Windows 7 here. Can you describe precisely what you do to achieve that?

>> can I likewise make use of the icon configured in a shortcut?

> That should definitely work.

How? The problem is that mintty sets its own icon, overriding any implicit icon cloned from the shortcut.
So how could mintty know whether it was called from a shortcut and retrieve the icon being configured there? Or should mintty not set its icon explicitly anymore? Not sure what implications that might have...

---
(getting off-topic, but anyway...)
>> Not necessarily. It's the shell that decides where to start, as a result of its profile processing.

> Yeah, and the Cygwin/MSYS2 shell always decide to start in the home directory by default. Or did I misunderstand?

It's not the shell, it's the profile. /etc/profile has a special case if CHERE_INVOKING is set, also everyone is free to change their own /etc/profile.

---
(icon should be set to the Windows path)
>> I'm adding a hint to the manual page (rather than converting) and a complete example.

> That doesn't sound too good. I find it nice that we can specify icons like /msys2.ico. Will Windows-style relative paths work at least?

They will, as well as cygwin paths, in any individual invocation, just not for configuring taskbar properties. As the syntax for that is ugly anyway, and not intended for case-by-case use, I think that's OK.

elieux:
**ad shortcuts - start in**) I'm on Windows Server 2012R2 English.

*pin from shortcut*) I drag'n'drop `C:\msys64\usr\bin\mintty.exe` using the right mouse button to `C:\` and choose "Create shortcut". Right click on `C:\mintty - Shortcut.lnk` and "Start in" is `C:\msys64\usr\bin`. I change it to `C:\msys64`. Right click on shortcut and "Pin to Taskbar". Right click on new icon on the taskbar, right click on "mintty - Shortcut", "Properties". "Start in" is `C:\msys64`.

*pin from process*) I run `C:/msys64/usr/bin/mintty.exe -i 'C:/msys64/msys2.exe' -o 'AppLaunchCmd=C:/msys64/msys2.exe' -o 'AppID=MSYS2.Shell.MSYSTEM.8' -o 'AppName=MSYS2 MSYSTEM Shell' --store-taskbar-properties -- /usr/bin/bash --login -i`, then right click the icon on the taskbar and "Pin this program to taskbar". Close the terminal and right click on the icon, right click on "MSYS2 MSYSTEM Shell", "Properties". "Start in" is `C:`.

Not sure how to fix this and [MSDN on AppUserModel properies](https://msdn.microsoft.com/en-us/library/windows/desktop/ff521732(v=vs.85).aspx) doesn't help.

**ad shortcuts - icon**) It seems I misunderstood your question. Can you rephrase the original question?

**ad `--cd` option**) I know about that option. The point was that
- by default, the shell goes to the home directory, so the proposed option has no effect
- if a person uses `CHERE_INVOKING`, they probably also know how to change the "Start in" property of the shortcut, so the proposed option is unnecessary

That applies to MSYS2 and Cygwin. As for GfW, I think the the `git-bash` wrapper does all the needed directory changing. For other applications, the option could of course be beneficial.

**ad icon path**) Okay then.

elieux:
> I run `C:/msys64/usr/bin/mintty.exe -i 'C:/msys64/msys2.exe' -o 'AppLaunchCmd=C:/msys64/msys2.exe' -o 'AppID=MSYS2.Shell.MSYSTEM.8' -o 'AppName=MSYS2 MSYSTEM Shell' --store-taskbar-properties -- /usr/bin/bash --login -i`, then right click the icon on the taskbar and "Pin this program to taskbar". Close the terminal and right click on the icon, right click on "MSYS2 MSYSTEM Shell", "Properties". "Start in" is `C:`.
> Not sure how to fix this and MSDN on AppUserModel properies doesn't help.

A-ha! This is just lazy Microsoft devs reinventing the wheel (i.e. path parsing). If I use backslashes, "Start in" shows up correct. `C:/msys64/usr/bin/mintty.exe -i 'C:\msys64\msys2.exe' -o 'AppLaunchCmd=C:\msys64\msys2.exe' -o 'AppID=MSYS2.Shell.MSYSTEM.8' -o 'AppName=MSYS2 MSYSTEM Shell' --store-taskbar-properties -- /usr/bin/bash --login -i`

mintty:
Pin from shortcut: confirmed, this works. But if the shortcut is started first and then the process is pinned, it fails, i.e. the shortcut parameters get lost, whether "Start in" or "Change Icon" - which was my other question.

mintty:
Released in 2.2.1.
Feel free to continue discussion here anyway.

elieux:
> Pin from shortcut: confirmed, this works. But if the shortcut is started first and then the process is pinned, it fails, i.e. the shortcut parameters get lost, whether "Start in" or "Change Icon" - which was my other question.

Right. I don't think the process can detect what shortcut was used to start it. Pinning a running process should be essentially the same as pinning the executable file (barring any AppUserModel manipulations).

FraGag:
Processes can detect if they were started from a shortcut, and from what shortcut. [`GetStartupInfo`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms683230%28v=vs.85%29.aspx) sets the `STARTF_TITLEISLINKNAME` flag in the `dwFlags` field of the [`STARTUPINFO`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms686331%28v=vs.85%29.aspx) structure if the process was started from a shortcut, and the `lpTitle` field contains the path of the shortcut.

mintty:
Thanks for the hint, but in my test, when starting from a shortcut, the LINK flag was not set.

mintty:
Withdrawing previous comment, the flag and shortcut file name can be retrieved. Now the icon name is more fiddly...

dscho:
> The point was that
> 
> * by default, the shell goes to the home directory, so the proposed option has no effect
> * if a person uses CHERE_INVOKING, they probably also know how to change the "Start in" property of the shortcut, so the proposed option is unnecessary
> 
> That applies to MSYS2 and Cygwin. As for GfW, I think the the git-bash wrapper does all the needed directory changing. For other applications, the option could of course be beneficial.

Unfortunately, you cannot set `CHERE_INVOKING` in a command-line invocation nor in a shortcut. That's why `git-bash.exe` introduced the `--cd=...` option and in the meantime it has learned *not* to mess with the current directory by default.

You might want to consider a similar solution in MSys2.

elieux:
I think that the new MSYS2 launcher will be fine in this regard (go to home directory when no arguments unless configured otherwise; always stay in current directory when a command is run), but it's possible it'll need to be implemented at some later time.

I think this ticket is overloaded. I suggest creating new ones for `--cd` and shortcut icon detection (and linking those here), if needed.

dscho:
It is really sad that the failure to work together on this issue caused so much troubles, not only for you, but now also for me. Really sad.

mintty:
Not sure what you mean to address with this obscure comment. I think we worked together well from both sides, trouble being caused by MS and their weird ideas. You aren't by any means pouting because I hadn't simply accepted your patch seamlessly?
Also the referred "failed to fork child process" isn't related to the taskbar issue discussed here. Moreover, both have meanwhile been fixed. With a new installation of Git for Windows I noticed it's still using a pretty old version of mintty – maybe that's causing trouble too?
🎅

mintty:
See also #493 for the referred "Failed to fork child process" problem. This was also discussed on the cygwin mailing list. It's actually a Windows bug for which cygwin has a workaround (not yet released). For Git you'll likely have to achieve a similar workaround in MSys.

dscho:
> I think we worked together well from both sides

That does not match my experience.

mintty:
I guess you agree that the involved Windows feature is a very obscure and also poorly documented one. After you worked a month on a working patch, you should accept that I needed a month too to understand the issue, ponder alternatives (as the suggested solution isn't really nice due to its irreversible nature) and test everything. And as you are involved in open source yourself you should also be aware that this month had to be distributed over some time. If you would raise an issue of similar weirdness with, say, MS, what response time would you expect, honestly?

mintty:
Release 2.2.3: Deriving icon from desktop shortcut.

PR #490, opened 2015-11-10
fix ECH parameter overflow bug.

ttdoda:
ECH parameter overflow bug was found in putty.
It seems that mintty has the same problem.

c.f. http://www.chiark.greenend.org.uk/~sgtatham/putty/wishlist/vuln-ech-overflow.html

### DIFF
--- a/src/termout.c
+++ b/src/termout.c
@@ -723,6 +723,8 @@ do_csi(uchar c)
   term_cursor *curs = &term.curs;
   int arg0 = term.csi_argv[0], arg1 = term.csi_argv[1];
   int arg0_def1 = arg0 ?: 1;  // first arg with default 1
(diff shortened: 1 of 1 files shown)


mintty:
Thanks; I guess I'd rather fix it like this:
```
   int arg0 = term.csi_argv[0], arg1 = term.csi_argv[1];
+  if (arg0 < 0)
+    arg0 = 0;
+  if (arg1 < 0)
+    arg1 = 0;
   int arg0_def1 = arg0 ?: 1;  // first arg with default 1
```
so to handle it like a signed positive value-range overflow.

mintty:
Added multiple guards.

mintty:
Fixed in 2.2.2.

PR #532, opened 2016-03-28
Update to build correctly on Cygwin 2.5

elieux:
Some functions, like `asprintf` and `ptsname`, are now available only after
requesting non-POSIX behavior, using `_GNU_SOURCE` and `_XOPEN_SOURCE`.

Include `<string.h>` in `ctrls.c` for `strdup`.

Cygwin 2.5.0 is currently in testing (unreleased). MSYS2 has already integrated these changes with the  2.5.0.17040.9d5d874-1 release.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -80,7 +80,7 @@ rc_srcs := $(wildcard *.rc)
 objs := $(c_srcs:.c=.o) $(rc_srcs:.rc=.o)
 bins := $(patsubst %.o,$(BINDIR)/%.o,$(objs))
 
-CFLAGS := -std=gnu99 -include std.h -Wall -Wextra -Wundef -Werror
+CFLAGS := -D_XOPEN_SOURCE -D_GNU_SOURCE -std=gnu99 -include std.h -Wall -Wextra -Wundef -Werror
 
 ifeq ($(shell VER=`$(CC) -dumpversion`; expr $${VER%.*} '>=' 4.5), 1)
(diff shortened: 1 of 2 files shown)


mperry2:
Are you sure that `_XOPEN_SOURCE` is needed? I was able to compile by adding `_GNU_SOURCE`.

mintty:
Thanks, I'll rather put the defines into std.h. I commited that with you as author.
string.h is already included from std.h

elieux:
@mintty, cool. I didn't know there was a `std.h`, but I don't see `std.h` being included in `ctrls.c`. (I know it's not important for the build at the moment so if you don't want to, I won't discuss this further.)

@mperry2, `_GNU_SOURCE` is enough, but adding both defines better expresses the intention IMO.

mintty:
See Makefile: -include std.h ; I don't know why it's included this way.
According to http://linux.die.net/man/3/ptsname, _XOPEN_SOURCE should be required for ptsname, so I included it just in case this will be changed to conform more strictly in cygwin in the future.

elieux:
Ah. Thanks for the [Makefile reference](https://github.com/mintty/mintty/blob/4b581a2/src/Makefile#L83), now I get it.

Deleting my branch.

mintty:
According to http://linux.die.net/man/7/feature_test_macros, _GNU_SOURCE implies _XOPEN_SOURCE, so removing the latter.

PR #571, opened 2016-07-30
Add SIXEL graphics support

saitoha:
@mintty Try if you like it.
https://youtu.be/qreXyqH8I1s

### DIFF
--- a/LICENSE.xterm
+++ b/LICENSE.xterm
@@ -1,0 +1,34 @@
+The helper function hls_to_rgb in src/sixel.c is imported from
+graphics.c in Xterm pl#310 originally written by Ross Combs.
+
+http://invisible-island.net/xterm/
+
+
+Copyright 2013,2014 by Ross Combs
+
+                        All Rights Reserved
+
+Permission is hereby granted, free of charge, to any person obtaining a
+copy of this software and associated documentation files (the
+"Software"), to deal in the Software without restriction, including
+without limitation the rights to use, copy, modify, merge, publish,
+distribute, sublicense, and/or sell copies of the Software, and to
+permit persons to whom the Software is furnished to do so, subject to
+the following conditions:
+
+The above copyright notice and this permission notice shall be included
+in all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+IN NO EVENT SHALL THE ABOVE LISTED COPYRIGHT HOLDER(S) BE LIABLE FOR ANY
+CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+Except as contained in this notice, the name(s) of the above copyright
+holders shall not be used in advertising or otherwise to promote the
+sale, use or other dealings in this Software without prior written
+authorization.
+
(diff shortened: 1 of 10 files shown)


mintty:
Gorgeous! Amazing how fast you could integrate Sixel support into mintty, and it even basically works with the scrollback buffer and the alternate screen as you demonstrate in the video. Thanks a lot. I would have liked to try it myself but was refraining from the effort...

There are some glitches still with the scrollback buffer; try Shift+PgUp/Dn, or searching with Alt+F3, and images will sometimes be placed in the wrong position. That might be even an acceptable glitch for a first release, though.

Why LICENSE.xterm? Why not LICENSE.sixel from libsixel?

How are colours handled? If I recall correctly, the sixel format is palette-based but with the option to tune the palette. That would mean that the sixel code should translate to true-color which mintty uses for output, right?

What about tools? Would you package them for cygwin once mintty is supporting it?
I vaguely remember that many years ago an `ls` clone was announced to output icon images for file types (not sure whether it was even sixel-based). Are you aware of any such tool? Might be a nice incentive to propagate the sixel feature in the cygwin community.

saitoha:
First, I apologize that I had not read mintty repo's policy. Should I create another issue?

>There are some glitches still with the scrollback buffer; try Shift+PgUp/Dn, or >searching with Alt+F3, and images will sometimes be placed in the wrong >position. 

Woah..., this patch is still buggy. It need to be improved continuously until this feature becomes stable.

> Why LICENSE.xterm? Why not LICENSE.sixel from libsixel?

OK, LICENSE.sixel is also needed. I'll add it.
LICENSE.xterm is needed only for the tiny function hls_to_rgb() in sixel.c.

>How are colours handled? If I recall correctly, the sixel format is palette-based >but with the option to tune the palette. That would mean that the sixel code should translate to true-color which mintty uses for output, right?

Yes, the implementation of sixel.c parses adaptive palette up to 256 color registers and convert it to RGB-based DIB section.

>What about tools? Would you package them for cygwin once mintty is supporting it?

I will propose cygwin port of libsixel and sixel support for w3m.
(Debian's w3m branch already has `-sixel` option)
Some existent packages that support sixel already ported to cygwin: 

gnuplot


ghostscript


netpbm


ImageMagick


>I vaguely remember that many years ago an ls clone was announced to output >icon images for file types (not sure whether it was even sixel-based). Are you >aware of any such tool? 

XMLterm?
ihttp://www.xml.com/pub/a/2000/06/07/xmlterm/

mintty:
> First, I apologize that I had not read mintty repo's policy. Should I create another issue?

Let's first sort out some open issues, then we'll see. If an existing pull request cannot be updated, a new one would be easy.

>>    There are some glitches still with the scrollback buffer; try Shift+PgUp/Dn, or >searching with Alt+F3, and images will sometimes be placed in the wrong >position.

> Woah..., this patch is still buggy. It need to be improved continuously until this feature becomes stable.

There'll be some time anyway, during my holidays :)

>    Why LICENSE.xterm? Why not LICENSE.sixel from libsixel?

> OK, LICENSE.sixel is also needed. I'll add it.
> LICENSE.xterm is needed only for the tiny function hls_to_rgb() in sixel.c.

I wouldn't like to have a confusing license file named "xterm". Couldn't it be simply be embedded in sixel.c like in xterm itself? And I'll mention it in the main LICENSE file. Or the function could be moved to a separate file, to distinguish licenses more clearly.

> XMLterm?

Looks like it, thanks.

There's one more bug apparently:
The cursor position after a sixel image is next to the right bottom of the image. I guess it should be at the beginning of the next line (checked with xterm, also mentioned in the xterm manual page for "sixel scrolling").

And we should also patch the Primary Device Attribute report in termout.c:
```
- static const char primary_da[] = "\e[?1;2;6;22c";
+ static const char primary_da[] = "\e[?1;2;4;6;22c";
```
Following xterm, one could also restrict sixel support to some "VT340" enabled mode, but I'd rather enable it always as that makes things unnecessarily complicated.

mintty:
xterm does not set a background (transparent background), which is not necessarily good if the terminal background is black. Your code sets white background which some people may not like for its brightness.
Would it be useful to have a choice? Does the sixel format offer an option already?

Where is actually your multiplt.dem file? I found one on the web but it shows only the lower right one of the 4 diagrams.

saitoha:
Moved to #572.

Issue #258, opened 2015-06-04
OSC 52 clipboard control (xterm)


GoogleCodeExporter:
```
small rfe, add support for xterm's clipboard OSC

paraphrasing from http://invisible-island.net/xterm/ctlseqs/ctlseqs.html

Operating System Controls
OSC 52 ; Pc ; Pd ST
OSC 52 ; Pc ; Pd BEL

-> Manipulate Selection Data.

The first parameter, Pc, is used to construct a list of selection parameters 
for clipboard, primary, select, or cut buffers (not important at all).  If the 
parameter is empty, xterm uses s 0 , to specify the configurable 
primary/clipboard selection and cut buffer 0. (Would be enough to always 
default to clipboard)

The second parameter, Pd, gives the selection data.  Normally this is a string 
encoded in base64.  The data becomes the new selection, which is then available 
for pasting by other applications. (This is important, base64 encoded data 
should be decoded and saved in the clipboard)

If the second parameter is a ? , xterm replies to the host with the selection 
data encoded using the same protocol.
```

Original issue reported on code.google.com by `ailin.ne...@gmail.com` on 8 Apr 2011 at 7:39

kangjianbin:
Hi mintty experts,

Is there still a chance for pull request #579? 
We need this because it is not a good user experience to use mouse to copy text if the line is wrapped.

With OSC52 support, we can also copy text in editors to client OS. Emacs supports this feature. And there are some configurations to make vim support it too.

This implementation has these limitations (or restrictions :-) ):
1. Only valid text is allowed to copy to clipboard.
2.  Doesn't allow server to read clipboard (ignores '?' command)
3. Selection parameters is ignored (always uses system clipboard)

That can avoid the security issues discussed above. 

If you still consider it to be dangerous, I can provide more restrictions:
1. Disable the feature by default unless user enables it in configuration (.minttyrc)
2.  Only allow clipboard selection parameter and ignore other parameters.

mintty:
I think with these restrictions, the feature can be integrated.
Here are some :) comments:

* As the second parameter is ignored, it should also work if it's empty (like in xterm).
* What do you mean by "Only valid text is allowed"? I don't see any related restriction in the code.
* win_copy_text: I suggest to use cs__mbstowcs (with double "_") which does the proper allocation, then free the memory.
* Please test whether non-ASCII characters can be copied from a non-UTF-8 terminal session and pasted properly into a UTF-8 session.
[...]

PR #579, opened 2016-09-04
Add Osc52 clipboard support.

kangjianbin:
Hi mintty develops,

I implemented an osc52 clipboard support. 
Is it acceptable to merge into mainline?


mintty:
This has been discussed in #258 already. Please consider the security discussion there and provide a convincing reason to introduce this functionality; quickly browsing the OSC description, it seems to be tuned towards the X11 clipboard model (primary, secondary, etc), please also present how this should map to Windows.
Technically, as another merge is pending which affects some of the same files, this one cannot be merged, thus closing this issue now.

Issue #572, opened 2016-07-30
Add SIXEL graphics support


saitoha:
Moved from #571.

>>> Why LICENSE.xterm? Why not LICENSE.sixel from libsixel?

>> OK, LICENSE.sixel is also needed. I'll add it.
>> LICENSE.xterm is needed only for the tiny function hls_to_rgb() in sixel.c.

> I wouldn't like to have a confusing license file named "xterm". Couldn't it be simply be embedded in sixel.c like in xterm itself? And I'll mention it in the main LICENSE file. Or the function could be moved to a separate file, to distinguish licenses more clearly.

I moved hls_to_rgb() to new separated file (4aa0ccd) .

> There's one more bug apparently:
> The cursor position after a sixel image is next to the right bottom of the image. I guess it should > be at the beginning of the next line (checked with xterm, also mentioned in the xterm manual page for "sixel scrolling").

OK, I fixed in 1d3ab0a.

> And we should also patch the Primary Device Attribute report in termout.c:
>
> - static const char primary_da[] = "\e[?1;2;6;22c";
> + static const char primary_da[] = "\e[?1;2;4;6;22c";

Fixed by aece93d.

saitoha:
>> There's one more bug apparently:
>> The cursor position after a sixel image is next to the right bottom of the image. I guess it should
>> be at the beginning of the next line (checked with xterm, also mentioned in the xterm manual page for "sixel scrolling").

> OK, I fixed in 1d3ab0a.

Sorry, amend by 90b744f.

mintty:
Still not like xterm (which moves to the horizontal beginning of the line, not of the image),
but note http://vt100.net/docs/vt3xx-gp/chapter14.html :
> When sixel display mode is enabled, the sixel active position begins at the upper-left corner of the ANSI text active position. Scrolling occurs when the sixel active position reaches the bottom margin of the graphics page. When sixel mode is exited, the text cursor is set to the current sixel cursor position.

So actually your previous approach was the original one and xterm does not implement it correctly.
I guess we should follow DEC here rather than xterm.
This applies to "Sixel Scrolling Mode Enabled"; "Sixel Scrolling Mode Disabled" does not seem very useful to me.

mintty:
[...]

PR #581, opened 2016-09-06
Add initial support for SIXEL graphics feature(DECSIXEL, #571, #572)

saitoha:
This patch adds SIXEL graphics feature to mintty (related: #572, #572).
After discussion of #572, I added a small bug fix https://github.com/saitoha/mintty-sixel/commit/f93d17f3b3026a197cac40593605c78be091fa55. This PR includes it.
Thanks for a lot of review by @mintty.

### DIFF
--- a/src/sixel.c
+++ b/src/sixel.c
@@ -1,0 +1,637 @@
+// sixel.c (part of mintty)
+// originally written by kmiya@cluti (https://github.com/saitoha/sixel/blob/master/fromsixel.c)
+// Licensed under the terms of the GNU General Public License v3 or later.
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <ctype.h>   /* isdigit */
+#include <string.h>  /* memcpy */
+
+#include "sixel.h"
+#include "sixel_hls.h"
+#include "winpriv.h"
+
+#define SIXEL_RGB(r, g, b) (((r) << 16) + ((g) << 8) +  (b))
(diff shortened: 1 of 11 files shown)


mintty:
The rendered Sixel image is spoiled after changing to/from the alternate screen (e.g. using less or an editor). Sorry I didn't notice it before, it was already the case with the previous version but not with the version of 2nd Aug.

mintty:
@saitoha: This broke in commit https://github.com/saitoha/mintty-sixel/commit/3910883242bc3e6d0dc8940892e39963b858e391 "Workaround for scrollback = 0 environment" while the commit before it, "TATTR_SIXEL is no longer needed" still worked.

mintty:
... which was only a small change. Reverting it in the current version fixes the behaviour. What was "Workaround for scrollback = 0 environment" meant to achieve?

saitoha:
> What was "Workaround for scrollback = 0 environment" meant to achieve?

If you revert saitoha@3910883 and set 0 to "Scrollback lines" setting, SIXEL scrolling will not work.

out

> The rendered Sixel image is spoiled after changing to/from the alternate screen

Sorry, I haven't noticed that problem. It is caused by wrong management of term.virtuallines. This patch will fix it.
https://github.com/saitoha/mintty-sixel/commit/4d55cdc0743e792258db0928a8be3b08a2bf2e13

PR #586, opened 2016-09-11
term: Add osc52 clipboard support.

kangjianbin:
Hi mintty experts,
See discussion in #258.
The feature is disabled by default. Nedd set 'AllowSetSelection' to true in .minttyrc to enable it.
Works with UTF-8 and non UTF-8 terminal.
Some limitation:
1. Read from clipborad is not supported.
2. No more than 1536 can be copy into clipboard. (Limits by length of cmd_buf)
3. Always use system clipboard.

### DIFF
--- a/src/base64.c
+++ b/src/base64.c
@@ -1,3 +1,7 @@
+// base64.c (part of mintty)
+// Copyright 2016 Jianbin Kang
+// Licensed under the terms of the GNU General Public License v3 or later.
+
 #include <errno.h>
 #include <stdint.h>
 #include "base64.h"
--- a/src/base64.c
+++ b/src/base64.c
@@ -1,0 +1,272 @@
+#include <errno.h>
+#include <stdint.h>
+#include "base64.h"
+
+static const char base64_table[] = {
+  'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H',
+  'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P',
+  'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X',
+  'Y', 'Z', 'a', 'b', 'c', 'd', 'e', 'f',
+  'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n',
+  'o', 'p', 'q', 'r', 's', 't', 'u', 'v',
+  'w', 'x', 'y', 'z', '0', '1', '2', '3',
+  '4', '5', '6', '7', '8', '9', '+', '/',
+};
+
+#define INVALID_CHAR	(-1)
+
+static inline char encode(uint32_t v)
+{
+  return base64_table[v];
+}
+
+static inline int decode(char v)
+{
+  if (v >= 'A' && v <= 'Z') {
+    return v - 'A';
+  }
+  if (v >= 'a' && v <= 'z') {
+    return v - 'a' + 26;
+  }
+  if (v >= '0' && v <= '9') {
+    return v - '0' + 52;
+  }
+  if (v == '+') {
+    return 62;
+  }
+  if (v == '/') {
+    return 63;
+  }
+  return INVALID_CHAR;
+
+}
+
+int base64_encode(const char *input, int ilen, char *output, int olen)
+{
+  int calc_len = (ilen + 2) / 3 * 4;
+  int i = 0;
+
+  if (olen < calc_len) {
+    return B64_OVERFLOW;
+  }
+  while (ilen >= 3) {
+    uint32_t v = (((uint32_t)input[0]) << 16) +
+      (((uint32_t)input[1]) << 8) + input[2];
+    output[i] = encode(v >> 18);
+    output[i + 1] = encode((v >> 12) & 0x3f);
+    output[i + 2] = encode((v >> 6) & 0x3f);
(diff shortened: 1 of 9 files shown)


mintty:
> 2. No more than 1536 can be copy into clipboard. (Limits by length of cmd_buf)

The limitation is not good.

> The comment of cmd_buf is OSC or DCS string buffer and length, so why not use it as OSC52 clipboard buffer? OSC52 is also a OSC sequence thought it needs larger buffer.

You're right. Let's make cmd_buf grow dynamically, e.g. like clip_workbuf in termclip.c. I may find some time to do that, or you may.

> `    win_copy(text, 0, size + 1);`

Is it proper to copy the terminating NULL character into the clipboard? Will it be pasted back then?
Please check / confirm.

kangjianbin:
> You're right. Let's make cmd_buf grow dynamically, e.g. like clip_workbuf in termclip.c. I may find some time to do that, or you may.

Agree. It is better to make it dynamically. Seems that tmux also uses dynamically allocation. But I prefer to send another request to enable it :-)

> Is it proper to copy the terminating NULL character into the clipboard? Will it be pasted back then?
Please check / confirm.

Looks like we need +1 when calling win_copy. 'win_copy_title' does the same way. If we don't add 1, the last byte will be lost. Would you please help confirm that?

mintty:
>> Is it proper to copy the terminating NULL character into the clipboard? Will it be pasted back then?

> Looks like we need +1 when calling win_copy. 'win_copy_title' does the same way. If we don't add 1, the last byte will be lost. Would you please help confirm that?

Right, I was looking at term_copy, but there a null is already appended by get_selection.

kangjianbin:
Yes. So looks like the code is ok. 
But I find another bug: If server sends '\x00' character, the code may discard all characters after the null character. I will verify if xterm supports '\x00' in osc52.

kangjianbin:
xterm supports '\x00' character:   _echo -e "he\x00llo" |yank_ will cause xterm to copy 'he^@llo' into clipboard. 

Looks like the 'cs__mbxx' functions don't support string contains '\0'.

Is it ok to add a 'cb__mbntowcn' function?

kangjianbin:
There is also a limitation in Windows clipboard: https://msdn.microsoft.com/en-us/library/windows/desktop/ff729168(v=vs.85).aspx

CF_TEXT/CF_UNICODETEXT: 
>A null character signals the end of the data.

So we can't support null character in Windows and needn't modify the code.

mintty:
OK about NULL.
Didn't you intend to implement dynamically growing cmd_buf? Right now, I think it could overflow...
Another question: What's the source of the base64 code; any copyright on it?

kangjianbin:
>Didn't you intend to implement dynamically growing cmd_buf? Right now, I think it could overflow...

My first implementation is to specify the size in .minttyrc and allocate with 'malloc'. I think dynamically growing with 'realloc' would be better. What's your opinion?

There is no overflow. If the command is too long, in CMD_STRING -> otherwise branch, some bytes will be dropped.

>Another question: What's the source of the base64 code; any copyright on it?

I wrote the base64 decoder/encoder. So the copyright will be same with other files :-). Need I add a header?

mintty:
> My first implementation is to specify the size in .minttyrc and allocate with 'malloc'. I think dynamically growing with 'realloc' would be better. What's your opinion?

Dynamically is desirable. A configuration option for such a limit is very archaic.

> I wrote the base64 decoder/encoder. So the copyright will be same with other files :-). Need I add a header?

No, as you like.

mintty:
I think the feature looks fine now but the handling of dynamic buffer growing looks a bit over-engineered for my taste. As term_release is only ever called in the event of premature killing of the application, we don't need it. Also, as the buffer only grows and is never shortened again, it can as well be handled locally. The nested init functions only create dependencies that I'd rather avoid.

And I'd suggest smaller initial size and steps, e.g. 128 (like in clip_addchar). There is also a realloc abstraction in mintty called renewn, which could be used (just for consistent style).

Furthermore, I need to check the logic of DCS_PASSTHROUGH where one level of abstract is skipped.

kangjianbin:
If term_release is necessary, then we needn't term_init either. I will initialize cmd_buf when term_write is called.

>Furthermore, I need to check the logic of DCS_PASSTHROUGH where one level of abstract is skipped.

What does it mean?

mintty:
I guess term_reset would be a proper place.

term_cmd_buf_push is called from term_push_cmd except after `when DCS_PASSTHROUGH:` where it is called directly.

kangjianbin:
DCS_PASSTHROUGH is different: if cmd_buf isn't larger enough, we can send the data first to purge the buffer. So we needn't discard any character.
I will change term_push_cmd to return a boolean value so that DCS_PASSTHROUGHT can call this function directly.
Another question: Which style mintty prefers: split term_push_cmd to small functions or merge the term_cmd_buf_xxx functions into term_push_cmd?

mintty:
> Which style mintty prefers: split term_push_cmd to small functions or merge the term_cmd_buf_xxx functions into term_push_cmd?

All-in-one is preferrable if there is no need to call the auxiliary function directly. If the abstraction still helps, you can also nest it as an inner function (often used e.g. in wininput.c).

mintty:
Released in 2.6.1.

PR #588, opened 2016-09-12
Fix per-monitor scaling of titlebars

jdmansour:
This fixes, at least for me, the scaling of titlebars when using multiple monitors with different DPI.

Everything works as expected, except for context menus which are still too big on the lower-DPI monitor.

One strange issue is that Windows sends a `WM_SIZE` message *during* the call to `EnableNonClientDpiScaling`. At this time, the window is not fully constructed (`WM_CREATE` has not been sent yet), but `win_adapt_term_size` tries to call `GetClientRect`, gets invalid data, and the program crashes. The `resizing = true;`, `resizing = false` code makes mintty ignore `WM_SIZE` during this phase.

Screenshot showing mintty before and after, on my regular-DPI external monitor (the laptop's screen is high-DPI):

titlebar_scaling

### DIFF
--- a/src/winmain.c
+++ b/src/winmain.c
@@ -1177,8 +1177,9 @@ static struct {
   switch (message) {
     when WM_NCCREATE:
       if (pEnableNonClientDpiScaling) {
-        CREATESTRUCT * csp = (CREATESTRUCT *)lp;
-        BOOL res = pEnableNonClientDpiScaling(csp->hwndParent);
+        resizing = true;
+        BOOL res = pEnableNonClientDpiScaling(wnd);
+        resizing = false;
         (void)res;
 #ifdef debug_dpi
         uint err = GetLastError();


mintty:
I can only try later but I wonder how this could work; I actually tried the same thing before, but this is WM_NCCREATE is sent during CreateWindow, so the result of CreateWindow has not yet been assigned to wnd which is NULL, so the call simply failed for me...

jdmansour:
Weird, in my case `wnd` was valid, but `csp->hwndParent` was NULL. I'm running insider build 14905 if it makes a difference. Maybe they changed the API at some point so that the future window handle is passed to WM_NCCREATE?

mintty:
Resolved the mystery. I was confused about `wnd` which is actually the parameter of win_proc here.
Thanks a lot for fiddling this together.

mintty:
The context menu is anchored to another instance of Windows magic, CreatePopupMenu.
So someone will need to find out how to set the font size of such a menu... An infinite story of API design crap.

mintty:
Released 2.6.1.

mintty:
Note that failure of the context menu to scale properly is also mentioned elsewhere (http://stackoverflow.com/questions/36864894/scaling-the-non-client-area-title-bar-menu-bar-for-per-monitor-high-dpi-suppo, look for `CreatePopupMenu`). Hopefully MS will fix this.

Mrodent:
W10. Bought a new monitor (4k i.e. 3840 x 2160).  Adjusted the W10 "graphics magnification" to 125%.  But the Cygwin/Mintty console is behaving badly: gigantic **system font** being used, which appears to be a multiplied factor of the font used by other Windows apps. I can adjust the Mintty **text font**, currently using 5 pt, but this is still quite large.  Where does Mintty go to decide the font it should use for the title, "Options" dialog, etc.?

mintty:
Which Windows version do you run? This is a Windows design bug and (partial) fixes are currently only available in the developers preview releases. You'll have to wait a few weeks for the first release containing the new big monitor scaling handling.

Mrodent:
Thanks!  I found a "workaround" : adjusted the "graphics magnification" back to 100%, rebooted and then reset to 125%: problem solved! I obviously "confused" the Windows 10 OS and its "marvellously professional" configuration mechanisms by, er, adjusting things too much.  Will keep an eye out for the next releases...

mintty:
Apparently fixed in current general Windows 10 release.

misse:
I'm running mintty 2.7.5 with windows 10 and am still seeing this issue. I can't find any new updates in windows update and I have an empty .minttyrc

Is there something I need to do to get this working? I've just updated to 2.7.5 and I've had this issue for as long as I can remember.

mintty:
Which is your Windows release?

misse:
Windows 10 Pro

mintty:
Which version (the 5-digit number)?

mintty:
Or which build (the 4-digit number shown by winver)?

misse:
Winver shows build 10240

mintty:
The titlebar scaling issue is fixed in Windows version 1607 build 14939.

PR #637, opened 2017-03-01
Add Japanese translations

k-takata:
I created Japanese translations of mintty.

However, I found some issues when I translated.

* The width of the option dialogbox is not enough to show the translated messages.
I needed to increase the width. E.g.:

```diff
--- a/src/res.rc
+++ b/src/res.rc
@@ -6,7 +6,7 @@
 
 IDI_MAINICON ICON "../icon/terminal.ico"
 
-IDD_MAINBOX DIALOGEX DISCARDABLE 32, 8, 256, DIALOG_HEIGHT
+IDD_MAINBOX DIALOGEX DISCARDABLE 32, 8, 310, DIALOG_HEIGHT
 STYLE DS_SHELLFONT | DS_MODALFRAME | WS_POPUP | WS_CAPTION | WS_SYSMENU
 CAPTION DIALOG_TITLE
 FONT 8, DIALOG_FONT
```

* Help messages (`--help`) and error messages for command line options are not translated.
It seems that `finish_config()` should be called before using `_()`. E.g.:

```diff
--- a/src/winmain.c
+++ b/src/winmain.c
@@ -2283,8 +2283,10 @@ main(int argc, char *argv[])
           ;
         else if (sscanf(optarg, "%i,%i%1s", &cfg.x, &cfg.y, (char[2]){}) == 2)
           ;
-        else
+        else {
+          finish_config();
           option_error(_("Syntax error in position argument '%s'"), optarg);
+        }
       when 's':
         if (strcmp(optarg, "maxwidth") == 0)
           maxwidth = true;
@@ -2294,8 +2296,10 @@ main(int argc, char *argv[])
           ;
         else if (sscanf(optarg, "%ux%u%1s", &cfg.cols, &cfg.rows, (char[2]){}) == 2)
           ;
-        else
+        else {
+          finish_config();
           option_error(_("Syntax error in size argument '%s'"), optarg);
+        }
       when 't': set_arg_option("Title", optarg);
       when 'T':
         set_arg_option("Title", optarg);
@@ -2335,12 +2339,14 @@ main(int argc, char *argv[])
       when 'D':
         cfg.daemonize_always = true;
       when 'H': {
+        finish_config();
         char * helptext = asform("%s %s %s\n\n%s", _(usage), APPNAME, _(synopsis), _(help));
         show_info(helptext);
         free(helptext);
         return 0;
       }
       when 'V': {
+        finish_config();
         char * vertext =
           asform("%s\n%s\n%s\n%s\n", 
                  VERSION_TEXT, COPYRIGHT, LICENSE_TEXT, _(WARRANTY_TEXT));
@@ -2349,8 +2355,10 @@ main(int argc, char *argv[])
         return 0;
       }
       when '?':
+        finish_config();
         option_error(_("Unknown option '%s'"), optopt ? shortopt : longopt);
       when ':':
+        finish_config();
         option_error(_("Option '%s' requires an argument"),
                      longopt[1] == '-' ? longopt : shortopt);
     }
```

* The output of `mintty --help` includes `--Reportpos s|o`.
I think this is an old name and it should be `--Report s|o`.

This PR doesn't include them.

### DIFF
--- a/lang/ja.po
+++ b/lang/ja.po
@@ -1,0 +1,1111 @@
+# Japanese translations for mintty package
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the mintty package.
+# Ken Takata <kentkt@csc.jp>, 2017.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: mintty\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2017-02-24 18:22+0100\n"
+"PO-Revision-Date: 2017-02-27 18:11+0900\n"
+"Last-Translator: Ken Takata <kentkt@csc.jp>\n"
+"Language-Team: \n"
+"Language: ja\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+
+#: charset.c:212 charset.c:223
+msgid "(Default)"
+msgstr "(既定)"
+
+#: charset.c:234
+msgid "(OEM codepage)"
+msgstr "(OEMコードページ)"
+
+#: charset.c:238
+msgid "(ANSI codepage)"
+msgstr "(ANSIコードページ)"
+
+#: child.c:76
+msgid "There are no available terminals"
+msgstr "使用可能な端末がありません"
+
+#: child.c:143
+msgid "Error: Could not open log file"
+msgstr "エラー: ログファイルを開けません"
+
+#: child.c:215
+msgid "Error: Could not fork child process"
+msgstr "エラー: 子プロセスをforkできません"
+
+#: child.c:217
+msgid "DLL rebasing may be required; see 'rebaseall / rebase --help'"
+msgstr "DLLのリベースが必要です; 'rebaseall / rebase --help' を参照"
+
+#. __ %1$s: client command (e.g. shell) to be run; %2$s: error message
+#: child.c:295
+msgid "Failed to run '%s': %s"
+msgstr "実行に失敗 '%s': %s"
+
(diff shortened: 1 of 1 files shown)


mintty:
Many thanks.
I've pulled in the ja.po file and will check your other comments.

mintty:
I'm looking for a convenient way to adapt the box width dynamically, rather than statically.

Untranslated text:
OK (may be OK)
Glass
About...
Enter
Wave
UI (user interface) [language]

mintty:
I've added a feature to scale the Options dialog width using a pseudo-entry in the localization file.
Added to ja.po:
msgid "100"
msgstr "121"

k-takata:
> OK (may be OK)

It's OK.

> Glass

I think this means Windows Vista's Aero Glass. "Aero Glass" is normally not translated in Japan. So this is OK.

> About...

This should be normally translated to `<Application Name>について...`. In this case `minttyについて...`. But there's not enough space for this.

> Enter

It's OK.

> Wave

I think this is "Wave file", then `Waveファイル(&W)` would be better, but not enough space.

> UI (user interface) [language]

This can be `ユーザーインターフェース言語`, but it is long. `UI言語` is enough.

> I've added a feature to scale the Options dialog width using a pseudo-entry in the localization file.

Thank you. I confirmed that it worked fine.

mintty:
Thanks a lot.

>>  About...

> This should be normally translated to <Application Name>について.... In this case minttyについて.... But there's not enough space for this.

What about "について" alone?

>>   Wave

> I think this is "Wave file", then Waveファイル(&W) would be better, but not enough space.

It could be translated like "Sound"; maybe I should change the label anyway.

k-takata:
> What about "について" alone?

It is very unnatural. "About" is better than it.

> It could be translated like "Sound"; maybe I should change the label anyway.

Then "Sound" can be translated to "音声".

mintty:
Some more tweaks, also for colour chooser geometry, released with 2.7.5.

mintty:

Issue #85, opened 2015-06-04
Searching the scrollback


GoogleCodeExporter:
```
Another Terminal.app feature that seems rather useful: the ability to
search the scrollback. Ctrl+/ is used to initiate that there.
```

Original issue reported on code.google.com by `andy.koppe` on 4 Apr 2009 at 7:52

kiwiz:
@peaches @mintty Is there interest in getting this into MinTTY? If so, I'd like to take a stab at it. :)

mintty:
Yes, any ideas how to do it? Blind search or Pop-up search form? Mark matches by selecting? Repeat search back/forward?

kiwiz:
Coolbeans. So I've been working on kiwiz/mintty which implements this (it's based on @peaches' changeset).

Differences between the two:
- Ctrl+Shift+H opens the search bar.
- Doesn't highlight when no results are found.
- Supports arbitrary length queries.
- TAB returns to the terminal. (This causes a beep that I haven't figured out yet)
- Should update correctly when the terminal updates.

Notes:
- `attr` was turned into a `unsigned long long` (from an `int`) which isn't great... Wasn't sure where else to stick the new bits.
- Wide character support is broken. Determining how many cells a character takes seems to be tricky. There's also the case where a wide char is on the final cell of a line and an extra space is inserted.
- `in_search` could be more efficient. It currently does a linear search over the results array for every cell on the screen.

Screenshot: http://i.imgur.com/BuNc2p5.png

mintty:
Hi Kai,
unfortunately, your code doesn't seem to compile here:
winsearch.c:145:7: error: initialization from incompatible pointer type [-Werror]
       .lpfnWndProc = search_proc,
       ^

Also, about the screenshot: Could it be possible to place the search input box above the window title, so the terminal screen area is not overlayed?
About attr: I'm currently reworking that anyway, so there would be more room for some bits if you need...
Further comments when I get the opportunity to try your patch, thanks.

kiwiz:
Hello!

I've pushed a change to the fork that might fix the problem. Let me know if that successfully compiles. Fwiw, I'm on Windows 8.1 and it compiles successfully under Cygwin's gcc 4.9.2.

Regarding the searchbar: I was trying to mimic where Chrome and iTerm2's. I like the effect because it's small and out of the way. That said, it does hide anything on the top right of the screen. What are your thoughts on making it a separate window or just reserving an entire line when the searchbar is opened?

[...]

PR #656, opened 2017-04-28
Unuseful variable in winsearch.c (searched).

paolo-sz:
It also introduces a delay on the first "next result" button push

### DIFF
--- a/src/term.c
+++ b/src/term.c
@@ -474,8 +474,10 @@ term_update_search(void)
     return;
   term.results.update_type = NO_UPDATE;
 
-  if (term.results.xquery_length == 0)
+  if (term.results.xquery_length == 0) {
+    term_clear_search();
     return;
+  }
 
   circbuf cb;
   // Allocate room for the circular buffer of termlines.
--- a/src/winsearch.c
+++ b/src/winsearch.c
@@ -6,7 +6,6 @@
 
 
 static bool search_initialised = false;
-static bool searched = false;
 static int prev_height = 0;
 
 static HWND search_wnd;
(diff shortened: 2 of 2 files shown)


mintty:
What delay? Please explain.
This affects code by @kiwiz (#85), so I'd like to give him opportunity to comment.

paolo-sz:
Just try to do the following:
- open the search bar
- write a single letter in the edit box in order to have multiple match on the screen (you will now see all the letters you typed highlighted and the first one highlighted as the first match)
- now push the right arrow on the search bar (the first match higilight should be supposed to move to the second match but this does not happens. only starting from the second push you can see the highlight to move...
This is the delay I'm talking about.

paolo-sz:
This better (my point of view) address the case which was managed by the searched variable (first match outside the current visible area).
Also, I feel the serch result it should be cleared if the search bar is cleared as well...

paolo-sz:
Or, even better, just scroll if current match is not visible otherwise change match and scroll (if necessary)

kiwiz:
@paolo-sz I feel like I had a reason for keeping the previous search if the search bar was cleared, but I don't remember why that was. Nonetheless, this change looks good to me!

paolo-sz:
@kiwiz First of all, I was wrong when I said that variable was not useful, sorry.
I think we fall into the same mistake by considering only one scenario of the two (maybe others) possible. This is the reason of my third commit.
About the previous search clears together with search bar, sorry, I'm not able to figure out when it could be useful too...

mintty:
I had the impression from comments above that patch/commit 3 replaces (part of?) patch 2. Not sure whether applying the merge would register all commits which is kind of confusing. Can we have a consolidated patch please? (And there seems to be an indentation quirk in line 29.)

paolo-sz:
Done (with indentation fix also).

Issue #501, opened 2015-12-07
Transparency Enhancement


jbisbee:
I'd like to ask for an enhancement to the background transparency.  In iTerm 2 on OSX they not only allowed for transpaency, but also allowed you to "blur" the transpancy as well.  This allowed for a great affect for coding windows as you could blur as little or as much as you wanted and the transparency would not affect code windows (vim/etc) when they stacked.

mintty:
Not sure what you mean; some screenshots maybe?
Also, do you have the Windows knowledge to implement such stuff?

jbisbee:
https://www.iterm2.com/documentation-preferences.html (CTRL+F "Blur").  I'll post screenshots from OSX later this evening.  And no, I presently don't have the knowledge, but it would be interesting to look into it.

jbisbee:
Maybe something. https://msdn.microsoft.com/en-us/library/windows/desktop/aa969537(v=vs.85).aspx#blurbehind

mintty:
I've added experimental blur-behind support. However, I don't see any positive effect in testing. On some systems, there is a very bad effect (big streaks over screen), on some there is no effect at all, on some the option also enforces glass background and not much more, where glass background implies blurring even without the option.
Please check (-o Blur=yes) with the current archive (Download ZIP button), but unless someone comes up with a working patch, I won't release the option.

jdmansour:
Hi all,
I've been working a bit on this, and I got transparency with blur-behind working on Windows 7 and 10 (and Windows 8 if you use [AeroGlass](http://www.glass8.eu/)) in my fork ([jdmansour/mintty](https://github.com/jdmansour/mintty)):

![Screenshot](https://raw.githubusercontent.com/jdmansour/mintty/glass/wiki/screenshot.png)

The default background color is partially transparent, while the foreground and the other background colors are opaque, as well as the title bar.

On Windows 10, I'm using the undocumented SetWindowCompositionAttribute (which Windows itself uses to blur behind the start menu), and on older Windows versions DwmExtendFrameIntoClientArea. The trick is to ensure that everything gets drawn with the correct alpha channel. It is not easy since plain GDI does not know about alpha.

I think this branch would need a bit of work before merging. The drawing code is a bit messy since I have to work around GDI's limitations. There are some small visual artifacts and it is not as performant as regular mintty. However it is already quite stable and pretty, and I've been using it daily for a while, so I thought I might share it.

mintty:
Thanks for this contribution, to be checked.
[...]

PR #661, opened 2017-05-16
Add support for windows 10

march1993:
Including the support for `SetWindowCompositionAttribute` function to enable blur behind the window.
Problem: it would become effective only after rebooting.

### DIFF
--- a/src/winmain.c
+++ b/src/winmain.c
@@ -114,13 +114,14 @@ trace_winsize(char * tag)
   printf("winsize[%s] @%d/%d %d %d cl %d %d + %d/%d\n", tag, (int)wr.left, (int)wr.top, (int)(wr.right - wr.left), (int)(wr.bottom - wr.top), (int)(cr.right - cr.left), (int)(cr.bottom - cr.top), extra_width, norm_extra_width);
 }
 #else
-#define trace_winsize(tag)	
+#define trace_winsize(tag)
 #endif
 
 
 static HRESULT (WINAPI * pDwmIsCompositionEnabled)(BOOL *) = 0;
 static HRESULT (WINAPI * pDwmExtendFrameIntoClientArea)(HWND, const MARGINS *) = 0;
 static HRESULT (WINAPI * pDwmEnableBlurBehindWindow)(HWND, void *) = 0;
+static HRESULT (WINAPI * pSetWindowCompositionAttribute)(HWND, void *) = 0;
 
 // Helper for loading a system library. Using LoadLibrary() directly is insecure
 // because Windows might be searching the current working directory first.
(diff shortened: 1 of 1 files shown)


mintty:
Thanks for your contribution. Compare to #501 with a similar patch.
However, I cannot currently reproduce the effect. Activating "glass" makes the background plain white here. I am aware that there used to be some Windows configuration conditions for glass mode (see manual page) but cannot replay that configuration either in a current Windows 10 version.
If you can provide more insights, especially a reproducible test case to actually make the effect visible, that would help to assess the patch.
(And by the way, I would prefer patches not to strip all line-ending spaces from the file, just to enhance overview...)

march1993:
I'm running with Windows 10 1703. It works somehow.

qq 20170517232720

Though many problems still remain. For example, text drawn by GDI doesn't have alpha channel.
I tried years ago to replace all TextOut with DrawThemeTextEx. But it's quite slow.
Another way is to use gdiplus, which brings even more problems :p

mintty:
Let me repeat more clearly: I do not see any difference by your patch. Everything looks like without the patch. If there is a difference in certain configuration cases, please provide details.

mintty:
@march1993 (and  @jdmansour):
It seems I have accidentally checked in this patch so it's now included in th 2.7.7 release, without mention in the changelog. Sorry for overlooking this. I will consider to leave it in as long as nobody complains.

And I will of course add credits for both of you (with similar contributions) in the commit history and on the credits wiki page provided that I can see a reproducible benefit of this in my environment. As said before, I don't see any difference now, so some setup instructions may help... I don't know whether this may be related to the hint in the mintty manual page, under the Transparency/Glass option. I also don't find this Windows setting anymore in Windows 10...

jdmansour:
Hi, cool, I'll try it out when I have some time. I don't remember if calling SetWindowCompositionAttribute was enough, though. It might work and you'll get everything transparent, including text (like right now when you choose e.g. 75% regular transparency), but in could also do nothing if the window has no alpha channel. @march1993, does the patch work for you?

One future improvement would be to add per-pixel-alpha, as in my fork. This way, text is 100% opaque, while the background is partially transparent and blurred. The way I implemented it is very messy and inefficient though. When I have some more time, I'll think about a better way.

Another improvement is that Microsoft has now announced "Fluent Design", which includes an official "Acrylic" blur effect. Unfortunately, this is only available in WPF apps. However, I think I've found out how it is implemented (as DirectComposition filters), and there has got to be a secret API to tell the windows DWM to apply a certain filter to a window. This is going to be a real challenge :-) but I think if it is possible, we will be able to just drop it in.

Guerra24:
Hi, the patch "works", not as expected but it does. The main issue is that blurs the entire window, including the window frame, and the text transparency is broken. For it to work: change to glass, save, close and open.

img

mintty:
Is that really how you want it to look? With also the text blurred (kind of, looking at the coloured-background lines)? And does the border look different after Save? For me it doesn't.

mintty:
Anyway, the way it looks in your screenshot is the same as it looked already before the patch (with -o Transparency=Glass).

Yamakuzure:
Well, for me it looks like the following screenshot on Windows 10 Pro. The screenshot is from today.

mintty_glass_01

(At least there is a glassy edge around the window, where the shadow should've been...)

Unfortunately I have no idea how to fix this.

PR #676, opened 2017-08-11
Visual bell: support alternative rendering of window-edges only

avih:
Flashing the whole window can be a bit jarring. This commit adds an
option for alternative rendering of flashing only the window's edges,
by setting BellFlashIsEdge=yes (no GUI, default is no).

---

I wasn't sure if `docs/mintty.1.html` should be edited manually, but it wasn't generated after modifying `docs/mintty.1` and `make`, so I ended up editing the html file manually.

I don't feel strongly about the names or the docs so I don't mind changing stuff if there are better suggestions, but the feature itself is a useful IMHO.

### DIFF
--- a/docs/mintty.1
+++ b/docs/mintty.1
@@ -1306,6 +1306,10 @@ The sound is also played when it is changed.
 \(en \fBFlash\fP (BellFlash=no): Briefly invert the foreground and background
 colours. (Corresponds to the xterm resource \fBvisualBell\fP.)
 .br
+\(en (BellFlashIsEdge=no): Alternative rendering of \fBFlash\fP
+- only flash the edges of the window, which may be less jarring than the whole
+window.
+.br
 \(en \fBHighlight in taskbar\fP (BellTaskbar=yes): Change the colour of mintty's
 taskbar entry if the mintty window is not active.
 (Corresponds to the xterm resource \fBbellIsUrgent\fP, 
--- a/docs/mintty.1
+++ b/docs/mintty.1
@@ -1306,9 +1306,8 @@ The sound is also played when it is changed.
 \(en \fBFlash\fP (BellFlash=no): Briefly invert the foreground and background
 colours. (Corresponds to the xterm resource \fBvisualBell\fP.)
 .br
-\(en (BellFlashIsEdge=no): Alternative rendering of \fBFlash\fP
-- only flash the edges of the window, which may be less jarring than the whole
-window.
+\(en (BellFlashStyle=0): Alternative rendering of \fBFlash\fP. 1 - only flash
+the edges of the window, which may be less jarring than the whole window.
 .br
 \(en \fBHighlight in taskbar\fP (BellTaskbar=yes): Change the colour of mintty's
 taskbar entry if the mintty window is not active.
--- a/docs/mintty.1.html
+++ b/docs/mintty.1.html
@@ -1,5 +1,5 @@
 <!-- Creator     : groff version 1.22.3 -->
-<!-- CreationDate: Fri Jul 28 16:07:34 2017 -->
+<!-- CreationDate: Fri Aug 11 17:21:53 2017 -->
 <!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN"
 "http://www.w3.org/TR/html4/loose.dtd">
 <html>
@@ -1596,6 +1596,9 @@ testing. The sound is also played when it is changed. <br>
 &ndash; <b>Flash</b> (BellFlash=no): Briefly invert the
 foreground and background colours. (Corresponds to the xterm
 resource <b>visualBell</b>.) <br>
+&ndash; (BellFlashIsEdge=no): Alternative rendering of
+<b>Flash</b> - only flash the edges of the window, which may
+be less jarring than the whole window. <br>
 &ndash; <b>Highlight in taskbar</b> (BellTaskbar=yes):
 Change the colour of mintty&rsquo;s taskbar entry if the
 mintty window is not active. (Corresponds to the xterm
--- a/docs/mintty.1.html
+++ b/docs/mintty.1.html
@@ -1,5 +1,5 @@
 <!-- Creator     : groff version 1.22.3 -->
-<!-- CreationDate: Fri Aug 11 17:21:53 2017 -->
+<!-- CreationDate: Fri Aug 25 16:36:09 2017 -->
 <!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN"
 "http://www.w3.org/TR/html4/loose.dtd">
 <html>
@@ -1596,9 +1596,9 @@ testing. The sound is also played when it is changed. <br>
 &ndash; <b>Flash</b> (BellFlash=no): Briefly invert the
 foreground and background colours. (Corresponds to the xterm
 resource <b>visualBell</b>.) <br>
-&ndash; (BellFlashIsEdge=no): Alternative rendering of
-<b>Flash</b> - only flash the edges of the window, which may
-be less jarring than the whole window. <br>
+&ndash; (BellFlashStyle=0): Alternative rendering of
+<b>Flash</b>. 1 - only flash the edges of the window, which
+may be less jarring than the whole window. <br>
 &ndash; <b>Highlight in taskbar</b> (BellTaskbar=yes):
 Change the colour of mintty&rsquo;s taskbar entry if the
 mintty window is not active. (Corresponds to the xterm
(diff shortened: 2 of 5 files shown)


avih:
So I deleted `docs/mintty.1.html` and then `make html` which seems to have generated it from `mintty.1` (ended in some errors, which I think are unrelated - the html modification looks good to me).

Updated the PR with the generated html file.

mintty:
Thanks, I'll check that after holiday time.

avih:
FWIW, it's also possible to use something like integer `BellFlashStyle` (instead of bool `BellFlashIsEdge`) where 0 (default) is full screen flash, 1 is edges, and it will still be open for more styles in the future.

I experimented with sides-only or top-bottom only or even bottom only (like tmux) - none is bad, but the full edges seemed to me a useful conservative option.

There's also a possibility to use more graphical visual bell in the future, maybe like iterm2 which can flash a graphical bell icon at the middle of the terminal. Not for this PR, but `BellFlashStyle` enables extensibility here.

mintty:
I've tried to use Windows border flash, FlashWindow, but that's very moderate and hardly noticeable.
One alternative might be to flash the whole area but with a more moderate variation of background colour rather than reversing it (tattr.attr ^= 1 << ATTR_BGSHIFT; as a quick test case).
If you think border cell flashing is still a benefit (after considering this option), I will include all options.

If you want your part pulled, please use cfg.bell_flash_style & 1 for border cell flashing, ... & 2 for full area flashing, please note that mintty does not use camel casing (so it's flash_char or flashchar), and patch term.c only, I will do the rest. Or I will make a cumulative patch if you agree.

avih:
> hardly noticeable

Hmm.. interesting. While indeed moderate, as that was my goal, it's unmissable for me. Maybe we have different color setups which makes its effect meaningfully different? (my default bg is 36,36,36 and default fg is 205,205,205).

> tattr.attr ^= 1 << ATTR_BGSHIFT;

I tried it, and while personally I still prefer the border, I agree it's also nice and less harsh that full screen inverse. However, apparently it does nothing when the bgcolor is fully black (0,0,0), so maybe it would need some more logic to handle it.

> If you want your part pulled, please use cfg.bell_flash_style & 1 for border cell flashing, ... & 2 for full area flashing, please note that mintty does not use camel casing (so it's flash_char or flashchar), and patch term.c only

What would it do for 0? invalid? disable BellFlash? or does it replace BellFlash (and disable on 0) ? Also, if I patch only term.c with cfg.bell_flash_style & 1 etc then it won't compile anymore without also modifying config.c/h .

For now, I only modified flashChar to flashchar (forced pushed the PR with this change only).

I don't mind at all if you take it from here by modifying the commit or adding more etc, or, alternatively, I'll do it - but I'll need replies for my questions above first.

Either way it would be trivial, so maybe better if you just take it from here and avoid communication pitfalls? :)

avih:
I added few commits:
- Make it bitwise style like you requested.
- An unrelated improvement (visible over selection).
- Experimental styles of modifying the background.

I'm very uncomfortable with the last one. Unless you have a sane way to modify the background such that it's always visibly different and non-hacky, I don't mind adding it. Otherwise, my attempts are not nice and I'd advise against them.

My suggestion (which I can do following your approval) is to squash `use bitwise BellFlashStyle` with the original commit, keep `be visible also over selection` as a separate commit, and remove `experimental alt-background rendering`.

Thoughts?

mintty:
Thanks for the update. I had in mind to use `brighten` as well, calling it directly, after converting colour to true colour value (like in the simplified form as used in wintext from line 1606, applied to bg instead of fg).
How about that?
And then (if it turns out well), since your main aim was to get a more moderate flashing effect, I'd propose to apply it to the whole terminal screen (not the borders), and unconditionally (replacing the old reverse), so no option would be required.

avih:
> and unconditionally (replacing the old reverse), so no option would be required.

Even though personally I don't use full-screen reverse, I think many would object to removing it (i.e. backlash after it's removed), since it's the most common type of visual-bell in other terminals as well. Also, supporting different visual bell styles happens to be relatively non intrusive to the code IMHO, so if it's easy to support different subjective preferences, why not?

FWIW, while I did play with background styles too, I personally still prefer the edge-inverse style which I originally suggested, and your suggestion doesn't include it either :/

As for using `brighten(...)`, other than the technicality where currently it's inaccessible at `term.c` (trivial to modify, but still), as I noted at the big comment at commit `experimental alt-background rendering`, I don't think it's logically within `term_paint(..)` mandate to modify explicit colors.

How about my suggestion my previous post? These would be trivial commits to add (support edge reverse as bitwise style), and then you can also add your `brighten` style as an additional option? :)

mintty:
I've uploaded my approach for your evaluation. I think the effect is good, and it's more generic this way.
We can still add your edge-flash, reverse-flash, and frame-flash as a set of styles, as you advocated.
The current upload is just for your evaluation.

avih:
> The current upload is just for your evaluation.

It's on master though ;) (i.e. I expected it to be on a branch for evaluation).

Anyway, I tested it and it seems to work, also with bg of 0,0,0. I can't say I follow the logic 100% - especially with DIM and that `0xFEFEFEFE` value, but it seems to be a similar approach, though more elaborate, to this https://github.com/mintty/mintty/pull/676/commits/37ff2dca1ad43604d23f8fa52040d867c22d0cc4#diff-366ece1c6e8026aed5df9c6ae82c3967R1188 , and with the same caveats: a bit too much logic for my taste, and IMHO out of place for `term_paint(...)`.

Style wise, and this is completely subjective, it's a bit too bright for my taste (that's also the reason I didn't use `brighten` and instead re-implemented my own with different offsets - I thought for this use case `brighten` modifies it more than I liked).

But it appears to be working, and it's your project :)

mintty:
> It's on master though ;) (i.e. I expected it to be on a branch for evaluation).

Yeah, I hate branching :/

> Anyway, I tested it and it seems to work, also with bg of 0,0,0.

Fine.

> I can't say I follow the logic 100% - especially with DIM and that 0xFEFEFEFE value, but it seems to be a similar approach, though more elaborate, to this 37ff2dc#diff-366ece1c6e8026aed5df9c6ae82c3967R1188 , and with the same caveats: a bit too much logic for my taste, and IMHO out of place for term_paint(...).

Well, it reuses what mintty already has, and I spent a lot of effort to tweak `brighten` to an acceptable result. For this reason this reuse is more convincing for me.

> Style wise, and this is completely subjective, it's a bit too bright for my taste (that's also the reason I didn't use brighten and instead re-implemented my own with different offsets - I thought for this use case brighten modifies it more than I liked).

I didn't notice a significant difference, maybe I should extend the flash length for testing (or maybe anyway?). If you wish to suggest a tweak and can achieve that by parameterising `brighten`, I will consider it.

mintty:
@avih, actually there's also a problem with double-width characters, which when located in the outer cells (the flashing terminal border cells) would likely yield a ragged flash pattern. To avoid such complication, I would again prefer to not have the border flash style, if you don't mind. I think the moderate full flash is a nice change and maybe sufficient.

mintty:
I've uploaded mode and configuration now as you proposed (--author=avih) but as said before, I'd rather remove the border mode at least from the manual page.

avih:
> actually there's also a problem with double-width characters

Personally I don't have double-width chars use cases, and indeed I haven't considered it, but I do see how it could have an effect.

> I would again prefer to not have the border flash style, if you don't mind ... at least from the manual page.

It's fully your decision to make. My goal with the PR was to make it available to others too because I thought that if I found it useful, others might too. Even if it's not at the code base, I can always just patch it locally, like I've been doing before this PR. If you think it's not beneficial enough to have it at the code base... up to you.

Anyway, the code as it is now with your latest commits - works for me.

As for the docs, I'd advise to keep the docs in sync with the implementation, whichever route you choose to take. IMO hidden options can be is a bit of a slippery slope. What you could do, however, is mark it at the docs as experimental and possibly to be removed in the future. This way you don't "commit" to keep supporting it.

And a tiny nit about the 8 (reverse) style bit: initially I used it (just 8) and it had no effect, and I then realized it only has an effect in conjunction with 2 or 4, so it might be worth noting this at the docs.

I also tried the 1 style value but couldn't notice any effect at the window frame, and I also didn't see its implementation at the code. I'm guessing you intend to add the implementation a bit later?

avih:
Current master has the feature this PR wanted to add, and more. Thanks! :)

Shall we close it?

mintty:
As you created the issue, you can close it, and thanks for asking; but usually I have the habit to close issues after the next release, to make clear a change is generally available.

avih:
> usually I have the habit to close issues after the next release

Sounds good to me, I'll leave it for you to close it.

Thanks again for the feature and for maintaining mintty in general!

mintty:
Released 2.8.0.

PR #679, opened 2017-08-16
Spanish translation file

jsanjose:
This is the Spanish translation file for mintty 2.7.9

Please, consider to add it to future releases. I will try to maintain it updated.


mintty:

PR #687, opened 2017-09-16
Add Simplified Chinese translation file

goreliu:
Add zh_CN.po


mintty:

PR #706, opened 2017-11-01
std.h/std.c: added the ARGZ_INTERNAL_IMPL macro for non-cygwin targets.

midipix:
At the present, mintty defines CYGWIN_VERSION_API_MINOR as 1007 for non-cygwin targets. While this measure solves many of the portability issues, it does not account for targets which do not provide (the non-standard) argz.h. The current patch retains the default behavior for cygwin builds while allowing also non-cygwin builds to succeed.

### DIFF
--- a/src/std.c
+++ b/src/std.c
@@ -86,7 +86,7 @@ int iswspace(wint_t wc) { return wc < 0x100 && isspace(wc); }
 #endif
 
 
-#if CYGWIN_VERSION_API_MINOR < 91
+#if (CYGWIN_VERSION_API_MINOR < 91) || (ARGZ_INTERNAL_IMPL)
 
 /* Copyright (C) 2002 by Red Hat, Incorporated. All rights reserved.
  *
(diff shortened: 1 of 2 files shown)


mintty:
So does midipix by default define ARGZ_INTERNAL_IMPL?

midipix:
By default we do, but if possible, I'd rather not tie that to the target or the __midipix__ macro, since that might change in the future.

mintty:
Can we use
#ifndef _ARGZ_H_
in order to avoid inventing a new indicator?

midipix:
That won't work, I'm afraid, because ARGZ_H (or actually _ARGZ_H_) is defined in argz.h, whereas the purpose of ARGZ_INTERNAL_IMPL is to decide whether to #include argz.h in the first place. That being said, a better name for the above macro might be HAVE_ARGZ_H or something along those lines.

mintty:
Closing this as too specific. Any progress on midipix actually? Once I see a complete compilable midipix port, I will check how to best integrate that into the mintty source repository.

PR #707, opened 2017-11-02
charset.h, charset.c: midipix-specific code paths are now handled extternally

midipix:
The previous mintty maintainer anticipated midipix-specific code paths in charset.h and charset.c, and added framework-specific tests accordingly. At the present, these tests are not only wrong, but also stand in the way of the current approach to porting mintty to midipix (using an external build system and avoiding any and all hard-to-maintain, midipix-specific code paths in the upstream mintty source tree). The current patch removes the above tests from chatset.c and charset.h.

### DIFF
--- a/src/charset.c
+++ b/src/charset.c
@@ -640,7 +640,7 @@ cs_btowc_glyph(char c)
   return wc;
 }
 
-#if defined(__midipix__) || defined(debug_wcs)
+#if defined(debug_wcs)
 
 unsigned int
 wcslen(const wchar * s)
@@ -666,7 +666,7 @@ wcscmp(const wchar * s1, const wchar * s2)
 
 #endif
 
-#if CYGWIN_VERSION_API_MINOR < 74 || defined(__midipix__) || defined(debug_wcs)
+#if CYGWIN_VERSION_API_MINOR < 74 || defined(debug_wcs)
 // needed for MinGW MSYS
 
 wchar *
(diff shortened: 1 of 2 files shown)


mintty:
So how would these functions be implemented then?
I would prefer to host a complete all-platform repository, rather than having patches applied.
We could factor out the distinct functions into one separate file, if you prefer; and there's also one #ifdef in termout.c.

midipix:
We've created a directory called src/host/midipix where we currently have two wrapper headers, std.h and charset.h, and where we will be adding the source files containing the implementation of host-specific function. Once everything is working on our end, it will be possible to simply merge the host-specific files into the upstream mintty repository by overlaying the former onto the latter. Regarding termout.c, the logic there is actually correct, but it would still be nice to use a designated macro name for that test rather than the generic framework one.

The host-specific headers can be found under https://github.com/midipix-ports/mintty/tree/midipix/src/host/midipix. The comment in std.h is a mere explanation of why we're taking this approach, and why overriding of libc interfaces (wcslen and friends) via macros is necessary.

midipix:
An update: 16-bit variants of wide-character interfaces are now provided by the framework's libc, where they are protected by a feature-test macro. This removes the need to implement and/or maintain them inside of mintty. For the newly added framework header (bits/wchar_u16.h) and updated host-specific std.h, see:

https://github.com/midipix-project/mmglue/blob/main/arch/nt64/bits/wchar_u16.h
https://github.com/midipix-ports/mintty/blob/midipix/src/host/midipix/std.h#L64

mintty:
I hope that finally redefinitions like #define wcslen wcslen_u16 would be triggered generically e.g. by #define WINAPI_UTF16_DEFAULT 1, and that #defining lots of U16_LITERALS is only a temporary patch.
Looking forward to a stable version. Closing this (as referring to work in progress).

midipix:
Sounds excellent. We now already compile -Wall -Werror -Wextra -Wundef 26 out of mintty's 27 translation units (all the .c plus res.rc), and hope to have all the w32api definitions that are needed by winmain.c sometime tonight. The next steps will then be linking (against automatically generated import libraries for the various w32 system dll's) and testing. The whole thing should be working rather soon, and we'll keep you posted.

Issue #708, opened 2017-11-04
New window comes up with super-small font when Ctrl+Shift+N is used


vadimkantorov:
I'm using WSLtty.

When I open a new session with Alt+F2, the font size is the same as in the original window. If I use Ctrl+Shift+N, then the font size is much smaller. The laptop screen resolution is quite high, so super-small fonts are impossible to work with. And Alt+F2 requires finger squeezing to press actually Alt+Fn+F2.

Would also be super-nice if the new window respected the original window size (and position?) -
 especially when the graphic virtual tabs start working. Currently I have to maximize it every time manually.

Thanks!
![screen](https://user-images.githubusercontent.com/1041752/32408649-ec11c7bc-c19b-11e7-87b9-f738cf64055b.jpg)

vadimkantorov:
The font problem happens mostly when the original windows was full-screen.

mintty:
This is somehow related to font zooming on screen resizing (#204) but of course there is no resizing involved here from a user point of view.

mintty:
Released 2.8.2.

Issue #478, opened 2015-09-16
Bold if same colors


lexazloy:
If i using same color and lite color i have lite color as bold:
```
echo -ne "\e]4;1;#ffffff\a"
echo -ne "\e]4;9;#ffffff\a"

echo -e "Naked \e[1;31mpeople\e[0m dancing"
```

> Naked **people** dancing

But if colors not same work as expected:

```
echo -ne "\e]4;1;#ffffff\a"
echo -ne "\e]4;9;#fffffE\a" # here

echo -e "Naked \e[1;31mpeople\e[0m dancing"
```

> Naked people dancing

P.S. `Show bold as font` unchecked, `Show bold as colour` checked.

mintty:
This is related to #468 but not quite the same issue.

You have redefined ANSI colours 1 and 9 and they do in fact behave as expected:
```
echo -ne "\e[38;5;1m ANSI colour 1 \e[0m"
echo -ne "\e[38;5;9m ANSI colour 9 \e[0m"
```
Both show the same redefined colour.

The question here is whether the ANSI bold attribute (\e[1m) should (with your config options) always simply map ANSI colours 0..7 to 8..15, overriding the principle which I'm trying to defend in #468 not to support "display bold as non-bold". What actually happens is that overstriking is applied as a fall-back to enforce distinct bold display, i.e. mintty is not as you said showing "lite as bold" but rather "bold as bold" - which is more or less what **I** would expect. However, it is arguable whether explicit redefining of the "colour implementing bold" should make an exception here. I will consider this and also check how other terminals behave.

lexazloy:
Thank you for answer. I saw previous ticket but dont understand how similar (or not) issue is.

If you say it pretty expected situation then ok. I mean i will little change second color and it give what i expect. Not a big deal.

But reason why i start issue. I tested few terms and every show my "version of truth". :)

Here Xfce4 Terminal, Xterm and Mintty:
![](https://i.imgur.com/UMZ44Kn.png)

Lxterminal:
![](https://i.imgur.com/6Tnf4LM.png)

Gnome Terminal:
![](https://i.imgur.com/FC96iag.png)

mintty:
Changing policy; with the latest change, overstriked bold display is not enforced anymore if the colour that the bold attribute gets mapped to is explicitly redefined (and happens to be the same as the non-bold colour).
[...]

Issue #468, opened 2015-08-11
Disabling 'Bold as font' still renders bold


silverwind:
Seems like this option is no longer working, fonts are always shown bold regardless if 'bold as font' is enabled or not. I downgraded to 2.1.3 to confirm this is an issue with the new version.

mintty:
There has been no change about bold attributes between 2.1.3 and 2.1.4.
Bold cannot be disabled altogether since 2.0.1 (this used to be possible for a while before).
Please describe the issue more clearly. What do you mean by "fonts are shown bold"? What settings are active, what do you output?

silverwind:
I'm talking about the 1px "shadow" rendered around text using theANSI bold code. You can see this with

```console
echo -e "\033[1mbold\033[0m text"
````

![bold](https://cloud.githubusercontent.com/assets/115237/9231892/eac8c2d2-412a-11e5-80f6-05542a35b516.png)

The option 'Show bold as font' should remove this shadow, but it has no effect and text is still rendered bold even if it's disabled. Also sorry, I've meant 1.1.3 as the last known working version.

mintty:
Changed previous issue title "Unable to disable 'Bold as font' in 2.1.4".
The option "Show bold as font" select a **corresponding** bold font for bold display, if one exists. That is not the case for Monaco which comes in only one style. In that case "bold as font" is not applied. That was like this in 1.1.3 already. The only change since then is that previously the fall-back behaviour was to not display bold at all while now mintty makes sure that one kind of bold display is applied in any case, the fall-back being the overstriking boldened display.
This works as designed; intending to close the issue.

silverwind:
Ah. It's the fallback behavious which I'd like to disable. Could you maybe provide an option in `.minttyrc` to disable the fallback? I find it visually unpleasing as it kind of disturbs the ClearType rendering.

aethelberht:
> The option "Show bold as font" select a corresponding bold font for bold display, if one exists.

I'm confused: how is that relevant to the fact that fallback modifications are still being done when the "Show bold as font" option is off?

mintty:
I'm not convinced that the bold character attribute, which is being used on purpose by applications, should be suppressed by configuration.
As I had explained on the cygwin mailing list already (https://cygwin.com/ml/cygwin/2015-08/msg00122.html): "I don't see why "bold as non-bold" should be an option." (This had been the case for a few mintty versions only until 1.1.3/1.2 but not for earlier versions either.)
[...]

PR #710, opened 2017-11-07
Improve bold configuration, improve dim behavior

avih:
(let me know if you want to split this PR)

Initially I only wanted to improve the bold font/colour/shadow control. I did
end up doing that, and these are the last two commits.

This is how BoldAsFont and BoldAsColour affect ANSI/default colours:
```
AsFont  AsColour  Previously                Now
------  --------  ------------------------  --------------------------------
no      yes       both only colour          both only colour (same, default)
yes     no        both only thick           both only thick (same)
yes     yes       both thick, ANSI +colour  both thick+colour (new)
no      no        both only thick           both thick, ANSI +colour (xterm)

Two combinations are the same, xterm-like moved from yes/yes to no/no,
and we now have an additional behaviour where both are thick+colour.

We also have the new BoldThickenForcesShadow to choose thick = font/shadow,
which defaults to font and is unaffected by the GUI checkboxes.
```

However, before I got there, I realized that the BoldAsFont/BoldAsColour logic
is duplicated 5 times around the code, that BoldAsColour also affects the
dim behavior, that the dim/BoldAsColour logic is duplicated 3 times, and that
there are few more similar duplications.

I could have duplicated my changes to all these places, or modify just enough
for the normal terminal rendering to be affected, but I chose instead to unify
all those duplicates of attributes-to-colours such that it's easy to modify in
the future as well.

So the unification is the first 3 commits and, except for new bugs, these 3
commits should keep the behaviour identical(*).

While decoupling BoldAsColour from dim, I also modified the dim behavior
(more info at the commit message):
```
Dim now does two things unconditionally:
- Blend the cell's foreground with the terminal's default background.
- Blend the cell's background with the terminal's default background.
```

(*) almost identical. I noticed a bug when copying true-colour to rtf, and it
was a one line fix. Other than this fix, it should be identical.

### DIFF
--- a/src/term.c
+++ b/src/term.c
@@ -1183,34 +1183,7 @@ term_paint(void)
         if (cfg.bell_flash_style & FLASH_REVERSE)
           tattr.attr ^= ATTR_REVERSE;
         else {
-          colour_i bgi = (tattr.attr & ATTR_BGMASK) >> ATTR_BGSHIFT;
-          if (term.rvideo) {
-            if (bgi >= 256)
-              bgi ^= 2;
-          }
-          colour bg = bgi >= TRUE_COLOUR ? tattr.truebg : colours[bgi];
-
-          colour_i fgi = (tattr.attr & ATTR_FGMASK) >> ATTR_FGSHIFT;
-          if (term.rvideo) {
-            if (fgi >= 256)
-              fgi ^= 2;
-          }
-          if (tattr.attr & ATTR_BOLD && cfg.bold_as_colour) {
-            if (fgi < 8) {
-              fgi |= 8;
-            }
-            else if (fgi >= 256 && fgi != TRUE_COLOUR && !cfg.bold_as_font) {
-              fgi |= 1;
-            }
-          }
-          colour fg = fgi >= TRUE_COLOUR ? tattr.truefg : colours[fgi];
-          if (tattr.attr & ATTR_DIM) {
-            fg = (fg & 0xFEFEFEFE) >> 1;
-            if (!cfg.bold_as_colour || fgi >= 256)
-              fg += (bg & 0xFEFEFEFE) >> 1;
-          }
-
-          tattr.truebg = brighten(bg, fg);
+          tattr.truebg = apply_attr_colour(tattr, ACM_VBELL_BG).truebg;
           tattr.attr = (tattr.attr & ~ATTR_BGMASK) | (TRUE_COLOUR << ATTR_BGSHIFT);
         }
       }
--- a/src/term.h
+++ b/src/term.h
@@ -60,6 +60,9 @@ typedef enum {
   TRUE_COLOUR = 0x180
 } colour_i;
 
+// colour classes
+#define CCL_ANSI8(i) ((i) < 8)
+#define CCL_DEFAULT(i) ((i) >= FG_COLOUR_I && (i) <= BOLD_BG_COLOUR_I)
 
 /* Special Characters:
  * UCSWIDE is a special value used in the terminal data to signify
--- a/src/term.h
+++ b/src/term.h
@@ -157,8 +157,10 @@ enum {
   ATTR_DEFAULT = ATTR_DEFFG | ATTR_DEFBG,
 };
 
+typedef unsigned long long cattrflags;
+
 typedef struct {
-  unsigned long long attr;
+  cattrflags attr;
   uint truefg;
   uint truebg;
 } cattr;
--- a/src/termline.c
+++ b/src/termline.c
@@ -242,7 +242,7 @@ makeliteral_attr(struct buf *b, termchar *c)
   * ensures that attribute values remain 16-bit _unless_ the
   * user uses extended colour.
   */
-  unsigned long long attr = c->attr.attr & ~DATTR_MASK;
+  cattrflags attr = c->attr.attr & ~DATTR_MASK;
   uint truefg = c->attr.truefg;
   uint truebg = c->attr.truebg;
 
@@ -314,7 +314,7 @@ readliteral_chr(struct buf *buf, termchar *c, termline *unused(line))
 static void
 readliteral_attr(struct buf *b, termchar *c, termline *unused(line))
 {
-  unsigned long long attr;
+  cattrflags attr;
   uint fg = 0;
   uint bg = 0;
 
--- a/src/termout.c
+++ b/src/termout.c
@@ -1567,8 +1567,8 @@ do_dcs(void)
 
     otherwise:
       /* parser status initialization */
-      fg = win_get_colour(term.rvideo ? BG_COLOUR_I: FG_COLOUR_I);
-      bg = win_get_colour(term.rvideo ? FG_COLOUR_I: BG_COLOUR_I);
+      fg = win_get_colour(FG_COLOUR_I);
+      bg = win_get_colour(BG_COLOUR_I);
       if (!st) {
         st = term.imgs.parser_state = calloc(1, sizeof(sixel_state_t));
         sixel_parser_set_default_color(st);
@@ -1700,7 +1700,7 @@ do_colour_osc(bool has_index_arg, uint i, bool reset)
     child_printf("\e]%u;", term.cmd_num);
     if (has_index_arg)
       child_printf("%u;", i);
-    c = win_get_colour(i);
+    c = i < COLOUR_NUM ? colours[i] : 0;  // should not be affected by rvideo
     child_printf("rgb:%04x/%04x/%04x\e\\",
                  red(c) * 0x101, green(c) * 0x101, blue(c) * 0x101);
   }
--- a/src/termout.c
+++ b/src/termout.c
@@ -630,7 +630,7 @@ do_sgr(void)
           uchar arg_10 = term.csi_argv[i] - 10;
           if (arg_10 && *cfg.fontfams[arg_10].name) {
             attr.attr &= ~FONTFAM_MASK;
-            attr.attr |= (unsigned long long)arg_10 << ATTR_FONTFAM_SHIFT;
+            attr.attr |= (cattrflags)arg_10 << ATTR_FONTFAM_SHIFT;
           }
           else {
             if (!arg_10)
@@ -641,7 +641,7 @@ do_sgr(void)
         }
       when 12 ... 20:
         attr.attr &= ~FONTFAM_MASK;
-        attr.attr |= (unsigned long long)(term.csi_argv[i] - 10) << ATTR_FONTFAM_SHIFT;
+        attr.attr |= (cattrflags)(term.csi_argv[i] - 10) << ATTR_FONTFAM_SHIFT;
       //when 21: attr.attr &= ~ATTR_BOLD;
       when 21: attr.attr |= ATTR_DOUBLYUND;
       when 22: attr.attr &= ~(ATTR_BOLD | ATTR_DIM);
@@ -2037,7 +2037,7 @@ term_write(const char *buf, uint len)
           continue;
         }
 
-        unsigned long long asav = term.curs.attr.attr;
+        cattrflags asav = term.curs.attr.attr;
 
         // Everything else
         int width;
@@ -2094,7 +2094,7 @@ term_write(const char *buf, uint len)
                   0, 0, 0, 0, 0, 0
                 };
                 uchar dispcode = linedraw_code[wc - 0x60];
-                term.curs.attr.attr |= ((unsigned long long)dispcode) << ATTR_GRAPH_SHIFT;
+                term.curs.attr.attr |= ((cattrflags)dispcode) << ATTR_GRAPH_SHIFT;
               }
 #endif
               wc = dispwc;
@@ -2115,7 +2115,7 @@ term_write(const char *buf, uint len)
                   0x81, 0x82, 0, 0, 0x85, 0x86, 0x87  // sum segments
                 };
                 uchar dispcode = techdraw_code[c - 0x21];
-                term.curs.attr.attr |= ((unsigned long long)dispcode) << ATTR_GRAPH_SHIFT;
+                term.curs.attr.attr |= ((cattrflags)dispcode) << ATTR_GRAPH_SHIFT;
               }
             }
           when CSET_GBCHR:  // NRC United Kingdom
--- a/src/win.h
+++ b/src/win.h
@@ -92,4 +92,14 @@ extern wchar win_combine_chars(wchar bc, wchar cc);
 
 extern wchar win_linedraw_char(int i);
 
+typedef enum {
+    ACM_TERM = 1,        /* actual terminal rendering */
+    ACM_RTF_PALETTE = 2, /* winclip - rtf palette setup stage */
+    ACM_RTF_GEN = 4,     /* winclip - rtf generation stage */
+    ACM_SIMPLE = 8,      /* simplified (for truecolour(...) ) */
+    ACM_VBELL_BG = 16,   /* visual-bell background highlight */
+} attr_colour_mode;
+
+extern cattr apply_attr_colour(cattr a, attr_colour_mode mode);
+
 #endif
--- a/src/winclip.c
+++ b/src/winclip.c
@@ -201,6 +201,17 @@ win_open(wstring wpath)
   }
 }
 
+static cattrflags
+apply_attr_colour_rtf(cattrflags a, attr_colour_mode mode, int * pfgi, int * pbgi)
+{
+  cattr ca = apply_attr_colour((cattr){ .attr = a }, mode);
+  *pfgi = (ca.attr & ATTR_FGMASK) >> ATTR_FGSHIFT;
+  *pbgi = (ca.attr & ATTR_BGMASK) >> ATTR_BGSHIFT;
+  // In ACM_RTF_GEN mode, COLOUR_NUM communicates "no-colour" (-1)
+  if (*pfgi == COLOUR_NUM && mode == ACM_RTF_GEN) *pfgi = -1;
+  if (*pbgi == COLOUR_NUM && mode == ACM_RTF_GEN) *pbgi = -1;
+  return ca.attr;
+}
 
 void
 win_copy(const wchar *data, uint *attrs, int len)
@@ -278,33 +289,7 @@ win_copy(const wchar *data, uint *attrs, int len)
     */
     memset(palette, 0, sizeof(palette));
     for (int i = 0; i < (len - 1); i++) {
-      uint attr = attrs[i];
-      fgcolour = (attr & ATTR_FGMASK) >> ATTR_FGSHIFT;
-      bgcolour = (attr & ATTR_BGMASK) >> ATTR_BGSHIFT;
-
-      if (attr & ATTR_REVERSE) {
-        int tmpcolour = fgcolour;     /* Swap foreground and background */
-        fgcolour = bgcolour;
-        bgcolour = tmpcolour;
-      }
-
-      if ((attr & ATTR_BOLD) && cfg.bold_as_colour) {
-        if (fgcolour < 8)     /* ANSI colours */
-          fgcolour += 8;
-        else if (fgcolour >= 256 && !cfg.bold_as_font)  /* Default colours */
-          fgcolour |= 1;
-      }
-
-      if (attr & ATTR_BLINK) {
-        if (bgcolour < 8)     /* ANSI colours */
-          bgcolour += 8;
-        else if (bgcolour >= 256)     /* Default colours */
-          bgcolour |= 1;
-      }
-
-      if (attr & ATTR_INVISIBLE)
-        fgcolour = bgcolour;
-
+      apply_attr_colour_rtf(attrs[i], ACM_RTF_PALETTE, &fgcolour, &bgcolour);
       palette[fgcolour]++;
       palette[bgcolour]++;
     }
@@ -376,56 +361,12 @@ win_copy(const wchar *data, uint *attrs, int len)
        /*
         * Determine foreground and background colours
         */
-        fgcolour = (attr & ATTR_FGMASK) >> ATTR_FGSHIFT;
-        bgcolour = (attr & ATTR_BGMASK) >> ATTR_BGSHIFT;
-
-        if (attr & ATTR_REVERSE) {
-          int tmpcolour = fgcolour;     /* Swap foreground and background */
-          fgcolour = bgcolour;
-          bgcolour = tmpcolour;
-        }
-
-        if ((attr & ATTR_BOLD) && cfg.bold_as_colour) {
-          if (fgcolour < 8)     /* ANSI colours */
-            fgcolour += 8;
-          else if (fgcolour >= 256 && !cfg.bold_as_font)  /* Default colours */
-            fgcolour |= 1;
-        }
-
-        if (attr & ATTR_BLINK) {
-          if (bgcolour < 8)     /* ANSI colours */
-            bgcolour += 8;
-          else if (bgcolour >= 256)     /* Default colours */
-            bgcolour |= 1;
-        }
-
-        attrBold = cfg.bold_as_font ? (attr & ATTR_BOLD) : 0;
+        attr = apply_attr_colour_rtf(attr, ACM_RTF_GEN, &fgcolour, &bgcolour);
+        attrBold = attr & ATTR_BOLD;
         attrUnder = attr & ATTR_UNDER;
         attrItalic = attr & ATTR_ITALIC;
         attrStrikeout = attr & ATTR_STRIKEOUT;
         attrHidden = attr & ATTR_INVISIBLE;
-
-       /*
-        * Reverse video
-        *   o  If video isn't reversed, ignore colour attributes for default
-        *      foregound or background.
-        *   o  Special case where bolded text is displayed using the default
-        *      foregound and background colours - force to bolded RTF.
-        */
-        if (!(attr & ATTR_REVERSE)) {
-          if (bgcolour >= 256)  /* Default color */
-            bgcolour = -1;      /* No coloring */
-
-          if (fgcolour >= 256) {        /* Default colour */
-            if (cfg.bold_as_colour && (fgcolour & 1) && bgcolour == -1)
-              attrBold = ATTR_BOLD;     /* Emphasize text with bold attribute */
-
-            fgcolour = -1;      /* No coloring */
-          }
-        }
-
-        if (attr & ATTR_INVISIBLE)
-          fgcolour = bgcolour;
 
        /*
         * Write RTF text attributes
--- a/src/winclip.c
+++ b/src/winclip.c
@@ -207,9 +207,9 @@ apply_attr_colour_rtf(cattrflags a, attr_colour_mode mode, int * pfgi, int * pbg
   cattr ca = apply_attr_colour((cattr){ .attr = a }, mode);
   *pfgi = (ca.attr & ATTR_FGMASK) >> ATTR_FGSHIFT;
   *pbgi = (ca.attr & ATTR_BGMASK) >> ATTR_BGSHIFT;
-  // In ACM_RTF_GEN mode, COLOUR_NUM communicates "no-colour" (-1)
-  if (*pfgi == COLOUR_NUM && mode == ACM_RTF_GEN) *pfgi = -1;
-  if (*pbgi == COLOUR_NUM && mode == ACM_RTF_GEN) *pbgi = -1;
+  // true colour breaks both, and ACM_RTF_GEN needs COLOUR_NUM as -1 (no colour)
+  if (*pfgi >= COLOUR_NUM) *pfgi = (mode == ACM_RTF_GEN) ? -1 : FG_COLOUR_I;
+  if (*pbgi >= COLOUR_NUM) *pbgi = (mode == ACM_RTF_GEN) ? -1 : BG_COLOUR_I;
   return ca.attr;
 }
 
--- a/src/wintext.c
+++ b/src/wintext.c
@@ -1401,7 +1401,6 @@ cattr
 apply_attr_colour(cattr a, attr_colour_mode mode)
 {
   // indexed modifications
-  bool do_rvideo_i = mode & (ACM_TERM | ACM_SIMPLE | ACM_VBELL_BG);
   bool do_reverse_i = mode & (ACM_RTF_PALETTE | ACM_RTF_GEN);
   bool do_bold_i = mode & (ACM_TERM | ACM_RTF_PALETTE | ACM_RTF_GEN | ACM_SIMPLE | ACM_VBELL_BG);
   bool do_blink_i = mode & (ACM_TERM | ACM_RTF_PALETTE | ACM_RTF_GEN);
@@ -1411,13 +1410,6 @@ apply_attr_colour(cattr a, attr_colour_mode mode)
   colour_i fgi = (a.attr & ATTR_FGMASK) >> ATTR_FGSHIFT;
   colour_i bgi = (a.attr & ATTR_BGMASK) >> ATTR_BGSHIFT;
   a.attr &= ~(ATTR_FGMASK | ATTR_BGMASK);  // we'll refill it later
-
-  if (do_rvideo_i && term.rvideo) {
-    if (CCL_DEFAULT(fgi))
-      fgi ^= 2;     // (BOLD_)?FG_COLOUR_I <-> (BOLD_)?BG_COLOUR_I
-    if (CCL_DEFAULT(bgi))
-      bgi ^= 2;     // (BOLD_)?FG_COLOUR_I <-> (BOLD_)?BG_COLOUR_I
-  }
 
   if (do_reverse_i && (a.attr & ATTR_REVERSE)) {
     colour_i t = fgi; fgi = bgi; bgi = t;
@@ -2428,6 +2420,8 @@ static bool bold_colour_selected = false;
 
 colour win_get_colour(colour_i i)
 {
+  if (term.rvideo && CCL_DEFAULT(i))
+    return colours[i ^ 2];  // [BOLD]_FG_COLOUR_I  <-->  [BOLD]_BG_COLOUR_I
   return i < COLOUR_NUM ? colours[i] : 0;
 }
 
--- a/src/wintext.c
+++ b/src/wintext.c
@@ -1343,6 +1343,147 @@ text_out_end()
     ScriptStringFree(&ssa);
 }
 
+
+// applies bold as colour if required, returns true if still needs thickening
+static bool
+apply_bold_colour(colour_i *pfgi)
+{
+  // We use two bits to control colouring and thickening for three classes of
+  // colours: ANSI (0-7), default, and other (8-256, true). We also reserve one
+  // combination for xterm's default (thicken everything, ANSI is also coloured).
+  // - "other" is always thickened and never gets colouring.
+  // - when bold_as_colour=no: ANSI+defaut are only thickened.
+  // - when bold_as_colour=yes:
+  //   .. and bold_as_font=no:  ANSI+default are only coloured
+  //   .. and bold_as_font=yes: ANSI+default are thickened, ANSI also coloured (xterm)
+  if (cfg.bold_as_colour) {
+    if (CCL_ANSI8(*pfgi)) {
+      *pfgi |= 8;     // (BLACK|...|WHITE)_I -> BOLD_(BLACK|...|WHITE)_I
+      return cfg.bold_as_font;
+    }
+    if (CCL_DEFAULT(*pfgi) && !cfg.bold_as_font) {
+      *pfgi |= 1;     // (FG|BG)_COLOUR_I -> BOLD_(FG|BG)_COLOUR_I
+      return false;
+    }
+  }
+  return true;
+}
+
+// removes default colours if not reversed, returns true if still needs thickening
+static bool
+rtf_bold_decolour(cattrflags attr, colour_i * pfgi, colour_i * pbgi)
+{
+  bool bold_thickens = cfg.bold_as_font;  // all colours
+  // if not reverse:
+  // - if only bold_as_colour, ATTR_BOLD still thickens default fg on default bg
+  // - don't colour default fg/bg (caller interprets COLOUR_NUM as "no colour").
+  if (!(attr & ATTR_REVERSE)) {
+    if (cfg.bold_as_colour && CCL_DEFAULT(*pfgi) && CCL_DEFAULT(*pbgi))
+      bold_thickens = true;  // even if bold_as_font=no
+    if (CCL_DEFAULT(*pfgi))
+      *pfgi = COLOUR_NUM;  // no colouring
+    if (CCL_DEFAULT(*pbgi))
+      *pbgi = COLOUR_NUM;  // no colouring
+  }
+  return (attr & ATTR_BOLD) && bold_thickens;
+}
+
+// Applies attributes to the fg/bg colours and returns the new cattr.
+//
+// "mode" maps to arbitrary sets of "things to do". Mostly these are just
+// groups of attributes to handle, but it may also reflect custom needs.
+//
+// While {FG|BG}MASK and truefg/truebg might update, attributes remain the same.
+// The exception is bold which can be applied as colour and/or thickness,
+// so ATTR_BOLD bit will be turned off if further thickening should not happen.
+// Always returns true colour, except ACM_RTF* modes which only do the palette.
+cattr
+apply_attr_colour(cattr a, attr_colour_mode mode)
+{
+  // indexed modifications
+  bool do_rvideo_i = mode & (ACM_TERM | ACM_SIMPLE | ACM_VBELL_BG);
+  bool do_reverse_i = mode & (ACM_RTF_PALETTE | ACM_RTF_GEN);
+  bool do_bold_i = mode & (ACM_TERM | ACM_RTF_PALETTE | ACM_RTF_GEN | ACM_SIMPLE | ACM_VBELL_BG);
+  bool do_blink_i = mode & (ACM_TERM | ACM_RTF_PALETTE | ACM_RTF_GEN);
+  bool do_finalize_rtf_i = mode & (ACM_RTF_PALETTE | ACM_RTF_GEN);
+  bool do_rtf_bold_decolour_i = mode & (ACM_RTF_GEN);
+
+  colour_i fgi = (a.attr & ATTR_FGMASK) >> ATTR_FGSHIFT;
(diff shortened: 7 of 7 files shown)


avih:
I noticed that github sorted the commits not by their correct order, so I updated the dates of two commits and force-pushed. Now the commits appear in correct order.

avih:
> I could have duplicated my changes to all these places, or modify just enough
> for the normal terminal rendering to be affected

What I also could do, and did in one version, was to extract/unify only the bold and dim procedures into their own functions and use them from 8 (combined) places. However, unifying all the attributes-to-colors procedures seemed the more correct approach, even if (a lot) more work. I think it came out well.

mintty:
I appreciate your ambitious care about boldness handling (although in general I prefer to discuss feature changes in advance). I am aware that there was quite some refactoring potential in this area. This is due to the unfortunate fact that a lot of side effects occur with boldness handling and changes are delicate, which can be seen in issues #468 and #478 and the number of related changes from releases 2.0.1 to 2.4.1.
For example, overstrike boldness is not mentioned in your description. Also there are some font dependencies and according implicit modes. I will check this and other issues thoroughly. This may take some time.

avih:
> although in general I prefer to discuss feature changes in advance

Agreed, but I started small and got carried away into deeper refactoring. Feel free to reject it with no hard feelings if you feel a different approach would have been better.



> overstrike boldness is not mentioned in your description

I don't think the term "overstrike" is mentioned at the mintty sources. It is a term which xterm uses though, and as far as I can tell the mintty equivalent is "shadow", AKA "fake bold" i.e. rendering non-bold font few times to emulate bold font.

It is mentioned. The second-to-last commit separate overstrike control into its own config, such that BoldAsFont can have only one function: control thickening of bold-text-attribute.

In general, every rendering which was possible before is still possible now, but with (hopefully) more useful configuration mapping and possibilities.

> This may take some time.

Understandable. Ping me with any issue or question you might have.

avih:
Also, the biggest commit by far is `refactor: unify attributes-to-colours logic (almost no-op)`, and the first three commits (the refactor is one of those) are basically a no-op, other than unifying the duplicates. So it should be simpler to verify as basically the behavior remains the same.

The actual functional changes at the later commits are rather simple and their commits are quite small - thanks to the earlier refactoring which made it possible for them to be simple.

mintty:
The reordered commits count 6, rather than 7 before; is everything included?

I'm not sure whether dim attribute blending of background colour is desirable; it's a matter of taste, or interpretation of what "dim" means. Did you maintain the "too close" logic that ensures sufficient contrast after blending?

Would your "don't force shadow" policy allow to effectively disable bold mode? I had argued repeatedly against any "bold as non-bold" option in the issues quoted above and still hold that opinion (finally appeased it with the BoldColour setting).

I would hesitate to change the AsFont/AsColour impact matrix, for the sake of the two quoted issues. The special case of ANSI colour application is justified by the traditional definition of colours 8..15 as "bright" colours 0..7.

The BoldThickenForcesShadow setting would make the current fallback behaviour questionable which is however needed because some fonts do not support BoldAsFont.

avih:
> The reordered commits count 6, rather than 7 before; is everything included?

These are the commits I'm seeing at the PR page (the top one is the latest), are any of those missing for you?
```
* 6972daf bold font/colour: improved control
* 574afcf BoldAsFont: decouple "thicken font" from "force shadow when disabled"
* c6c7b0b truecolour(..): remove and use attr_apply_colour(..) directly
* c6807f9 dim: simplify, improve, decouple from BoldAsColour
* f56b4c2 unify term.rvideo handling at win_get_colour (no-op)
* e258f28 refactor: unify attributes-to-colours logic (almost no-op)
* 6d8190c add typedef unsigned long long for cattr.attr (no-op)

```

> Did you maintain the "too close" logic that ensures sufficient contrast after blending?

I did not and could not "maintain" it, as previously dim didn't have such "too close" logic. I could have introduced it though, but I didn't. See below.

> I'm not sure whether dim attribute blending of background colour is desirable; it's a matter of taste, or interpretation of what "dim" means.

While true, I think the interpretation at this PR is better than both xterm's and mintty's. For reference, here's how it looks under various conditions in xterm, mintty before, and mintty after, and you can judge which best fits a description of "dim". That being said, it is subjective and I don't feel strongly about it. My main goal was the improving the bold configurability, and I only touched dim because I wanted to decoupled it from BoldAsColour.

xterm: left side: normal, reverse. right side: dim, reverse+dim:
xterm-all-combos

Mintty: left side: normal, reverse. middle: current mintty dim, reverse+dim, right side: this PR dim, reverse+dim:
mintty-current-new-all-combos

> Would your "don't force shadow" policy allow to effectively disable bold mode?

Depending how you define "bold mode", but I don't think so. Previously BoldAsFont effectively controlled two things:
- whether or not to try real bold font before falling back to shadow.
- Whether or not to thicken the text at all (when BoldAsColour=yes).

Now these two functions are split into two independent configs. That's it. BoldAsFont maintains the role of "thicken or not", while the other one controls what "thicken" means: whether or not we're allowed to try real bold font before falling back to shadow.


> I had argued repeatedly against any "bold as non-bold"

It was impossible to have "bold as non bold" before, and it's still impossible. The bold attribute was and is always portrayed as either thickened text (real bold font or overstrike), or coloured, or both. This was and still is the case.


> I would hesitate to change the AsFont/AsColour impact matrix, for the sake of the two quoted issues

In general I agree as I, too, value backward compatibility. However:
- The modified matrix still allows everything which was possible before.
- The default remains with the same behavior.
- It also allows something which wasn't possible: both ANSI and default are both thickened and coloured.
- One could argue that the new suggested matrix is more logical, since other than when both are "no", it behaves in a completely logical way where each config controls independently one feature (thickening/colouring). The only combination which is "abused" is the one where both are "no" which is used to make it behave like xterm's default behavior.

> The BoldThickenForcesShadow setting would make the current fallback behaviour questionable which is however needed because some fonts do not support BoldAsFont.

I don't _think_ that's the case, but I might have missed some of the logic. However, as far as my analysis and experimentation go, nothing is being left out. However, if you could point to a specific combination of configuration or fonts availability, I could look into that with more scrutiny and provide more qualified reply.

mintty:
I would prefer to first check whether we can find, instead of
`refactor: unify attributes-to-colours logic (almost no-op)`
a patch
`refactor: unify attributes-to-colours logic (reproducible no-op)`

If you want to name the "unsigned long long" flags type, I'd prefer `cattrflags`.

avih:
> I would prefer to first check ... a patch `refactor: unify attributes-to-colours logic (reproducible no-op)`

Sure. I'll split `refactor: unify attributes-to-colours logic (almost no-op)` into two commits: one with the refactor which keeps the behavior, and another commit which fixes the issue of true colour rtf copy.

> I'd prefer cattrflags

Sounds fine to me.

We're keeping the bold discussion for now at the other commit, yes? I think we should, as the bold subject deserves its own space.

Also, I don't mind much about dim, as nothing uses dim in practice. While I do still think it's an improvement, as far as I'm concerned take it or leave it as you see fit. Unless of course you want to discuss it (judging by the bottom-middle screenshot, it seems to behave weird with reverse+dim of the default fg colour on non-default bg. You can experiment/reproduce using https://github.com/avih/sgrtab ).

mintty:
Dim background: thinking again, if there is an area of specific background, and some parts are dimmed, I don't think it would look nice to have the background changed in spots. So dimming should affect only the foreground.
Colour tuning: for both dim and bold, the automatic brightening algorithm has an option to reverse the direction of the colour change, so bold can actually be dimmer and dim can be brighter, in border cases, as a consequence of the "too close" logic that ensures sufficient contrast to keep the text readable. It's a trade-off.

avih:
> Dim background: ...

Considering the following:
- Bold is used all the time.
- The code could benefit a lot from refactoring/unifying of attributes-to-colours (IMHO).
- Brighten is good enough right now.
- Dim is really never used in practice.

Allow me to suggest this: let's take dim out of the equation for now, keep this PR only about the refactoring, and keep #714 only about bold. The discussion we've conducted so far about bold is not about code but rather about changes in config/behavior, and its conclusion could be applied with or without prior refactoring and unrelated to modifications of dim/brighten.

Until we get to the dim/brighten discussion, here's some facts/food for thought:
- Dim is effectively always about bringing colours closer - which is guaranteed to be possible and simple with any pair of RGB triplets (unless they're the same colour, which is not an issue for the semantics of "dim") - it only needs decisions on which colours we want to bring closer and by how much.
- "Intensifying" is about increasing contrast between two colours - a thing which is guaranteed to _not_ always be possible, for instance if one colour is pure white and the other is pure black. Additionally, it may have complexities such as "what does bigger contrast between green and red even mean?".
- Fact (trust me, it is): the current behavior of `brighten(c, against)` is that if you pick two random RGB triplets, it "intensifies" about 30-40% of the cases, and for the rest it mostly looks like "dim". Importantly, the current use cases for `brighten` are such that mostly the results do fit the use case well, but It clearly required careful tweaking, and the various debug `#ifdef`, and constants such as `22222`  and others are a testament of how delicate and maybe frustrating the tweaking was.
- The current `brighten` cannot be a generic "intensifier" (for instance, if we wanted to do bold-as-colour for true colours) because most of the time it does "dim" rather than "intensify".

Let me know if you want me to remove the dim/bold parts from this PR. Either way I won't touch them at this PR until we're done with the refactor parts.

avih:
This forced-pushed set of commits is identical to the original set, except:
- Uses `cattrflags` instead of `cattr_attr_type`.
- Splits the "almost no-op" big refactor commit into two commits: no-op big refactor, and a small copy-to-rtf bugfix.

avih:
This forced-pushed set of commits is identical to the original set, except:
- Uses cattrflags instead of cattr_attr_type.
- Splits the "almost no-op" big refactor commit into two commits: no-op big refactor, and a small copy-to-rtf bugfix.

mintty:
> But when fgi/bgi are true colour, it's an out-of-bounds access (the TRUE_COLOUR fgi/bgi value is bigger than COLOUR_NUM).

Oh, indeed! I didn't care too much here when I introduced true colour because generating true colour into RTF does not seem to be feasible, because all colours need to be defined first, so we would need two passes over the whole buffer to extract all used colour values first.

What's actually the way to merge selected commits from a pull request? That's not offered by github, I think.

avih:
> What's actually the way to merge selected commits from a pull request?

I don't think you can from the github interface, and obviously the bugfix can only be applied after the refactor, as without the refactor the fixup code would be different.

Just tell me which commits you wanna merge, and I'll remove the other commits from the PR. I do have backup and I can open a new PR with the other commit after we finish discussing the bold subject.

So I'm guessing you want to merge:
```
* add typedef unsigned long long for cattr.attr (no-op)
* refactor: unify attributes-to-colours logic (no-op)
* bugfix: avoid read/write out-of-bounds when copying true colour as rtf
```
Yes?

What about `unify term.rvideo handling at win_get_colour (no-op)` ?

avih:
> because generating true colour into RTF does not seem to be feasible, because all colours need to be defined first

This is actually very feasible. Turning true colour to (xterm) 256 colours is fast and simple and only few lines of code. And the rtf code already supports coping xterm 256 colours. Once we're done with the refactor and bold, remind me to send a PR to make copy-to-rtf work with true colours too.

avih:
In reply to https://github.com/mintty/mintty/pull/714#issuecomment-344440018 :

> brute-force expanding it ... honestly, it's not straight-forward due to the many mode-dependent variants of the folded functions.

Yeah, brute force would be hard. For your convenience, here's the list I made of what each of the locations did before the refactor, and which I used while verifying the refactor against the removed code (i.e. 1. verify that the list matches each place before the refactor, and 2. that after the refactor it matches each mode). You might find it helpful. I did:
```
//  -- original processing and ordering at the different clients --

// ACM_TERM
// - rvideo
// - bold
// - blink
//   --> true colour
// - dim
// - reverse (true colour)
// - invisible (true colour)
// - search/mark highlights
// - cursor colour

// ACM_CP_PALETTE (only indexed)
// - reverse (indexed)
// - bold
// - blink
// - invisible (indexed)

// ACM_CP_RTF (only idnexed)
// - reverse (indexed)
// - bold
// - blink
// - decolour (!reverse -> remove default colours, bold s/colour/font/ for default)
// - invisible (indexed)

// ACM_SIMPLE
// - rvideo
// - bold
//   --> true colour
// - dim
// - invisible

// ACM_VBELL
// - rvideo
// - bold
//   --> true colour
// - dim
// - brighten-bg
```

> I'd actually prefer a more moderate folding approach

We could always just put the bold application into its own function, and instead of this big refactor we'll just have five places which call the `apply_bold_colours(...)` function while keeping the rest of their attributes handling duplicates.

And ignore dim (but I still think dim should not depend on BoldAsColour. Even with the historical investigation I still don't get what it tried to do with it and why, eventhough i did note the changes it went through). Or I'll just leave dim for you to fix/solve/whatever ;) I guarantee no one would complain however dim behaves. And you can send the complaints to me ;)

The only thing which I really think would be terrible would be to duplicate the bold config changes at 5 places. The refactor was something I think could benefit the code greatly, but of course, ultimately it's your decision if you want it or not.

> Remember I had asked for some time, and I didn't mean some hours.

As I said before, completely understandable. Take as much time as you need.

> For the records, I've been maintaining mintty since summer 2015

I know it might not be obvious considering my sometimes-harsh words, but I actually think mintty is one of the best terminals out there, and I want it to stay that way. I've been using it extensively for some months and it's been an absolutely smooth sail. Other than the bellflash border thingy and the bold issue - both of which are really minor in the grand scheme of things, it's been absolutely enjoyable and transparent experience.

So thank you for maintaining it and doing such a good job at it :)

mintty:
I have 3 sets of comments, and then I guess we can merge a version of the first 3 patches (or, if you don’t care about any automatic github credits, I could `commit --author=avih` them); 
thanks, by the way, that you are impressively strictly following the mintty naming and layout conventions:)

Also, I performance-tested the patch and that’s OK.
(I was worried about the passing of cattr values in function parameters and results, because after I had expanded the now cattrflags type from unsigned long to unsigned long long a while ago, this came with a performance penalty of 10% on a 32 bit processor − which luckily nobody noticed −; no penalty on a 64 bit processor).

---

I found 3 things to be a bit obscure, or not intuitive:
*  usage of the reset_bold flag
*  the role of rtf_bold_decolour in the workflow
*  despite the CCL_ attributes, apply_attr_colour still tests `fgi >= 256` once: is that on purpose?

---

The refactoring patch is not completely a no-op (although the difference makes it more consistent);
issues:
*  apply_bold_colour, do_rvideo for cursor and selection colours
*  apply_shadow in main colour handling section

To demonstrate, add the following in wintext.c, behind the `#ifdef debug_underline section`:
```
#define debug_bold
#ifdef debug_bold
  if (xwidth > 1) {
    force_manual_underline = true;
    ul = 0x80E02020;
  }
#endif
```
Run `mintty -o BoldAsColour=1 -o BoldAsFont=0`
`echo -e "\e[?5h"; cat xbold`
then select the output

---

I have a few ideas how to simplify the refactored code (but we may proceed here after merging an initial version, as you prefer):
*  unfold rtf_bold_decolour into winclip.c, and perhaps unify ACM_RTF_ modes
*  unfold do_vbell_bg (and keep term.c: tattr.truebg = brighten(bg, fg);), 
   and save ACM_VBELL_BG
*  maybe unfold do_reverse and do_blink_i (and check whether blink should set bold/colour at all), 
   and save ACM_TERM
*  unfold apply_bold_colour to have changes at a glance, 
   and maybe get the reset_bold logic clearer?
*  implement mapping of true colour → palette for RTF export

mintty:
> cat xbold

Forgot to attach my test file:
[xbold.txt](https://github.com/mintty/mintty/files/1478598/xbold.txt)

avih:
> Also, I performance-tested the patch and that’s OK

Thanks. May I ask what's your performance testing procedure/setup? I do have good background in performance (both at the code and in testing), and I did take into account that there could be some performance impact due to passing around `cattr` rather than working on it locally in-place, but I did not find a good way to measure it (I also didn't try very hard, but nevertheless I did look around and couldn't find a simple way).

Correct me if I'm wrong, but I think It would only really matter for `ACM_TERM` as the rest are really not performance critical, but this one is, and if possible, I'd also prefer not to regress its performance.

However, code quality and correctness come before performance, and performance can be improved later by identifying actual bottlenecks and devising a strategy to reduce them.

Also considering all the other things which happen around this code, I did assess that the perf impact would be somewhere between unmeasurable to negligible to very small. To be honest, 10% actually sounds worse than I thought, so thanks for bringing it up.

I'm looking at the other parts of your comment now. The review asks good questions which indicate you examined it very well, which I can and do appreciate.

avih:
> ... I could commit --author=avih them ...

Actually, you've done it also at the bellflash PR, and I was a bit surprised. In general if these are trivial changes then you can just amend the commit - which doesn't change the author details.

In this case however, I'd prefer to implement all your comments myself and also verify they're in line with the rest of the code, which would leave you eventually to just merge them.


Specifically about your comments:

> usage of the reset_bold flag

Not sure what exactly you want clarified. I want to answer accurately, and I think it could be two things:
- Why is ATTR_BOLD being reset at all.
- Why is it using reset_bold to apply it later rather than resetting it right after calling apply_bold_colour.

The brief is that ATTR_BOLD after the call is set if and only if the caller should (if it wants to) also apply thickening (bold font or shadow) - this is similar to the role of apply_shadow at the original at the bold code, but semantically more generic as it covers both bold font and shadow (technically it's an identical condition, as with the current code, when bold_as_font=no these are the exact cases where it would apply shadow, but in general, it says whether any kind of further thickening should happen).

The reason that reset_bold is used to delay the reset of ATTR_BOLD rather than doing it right after calling apply_bold_colour is that the later rtf_bold_decolour needs to know whether ATTR_BOLD was set and actually completely ignores what apply_bold_colour thinks about further thickening, but if it was reset earlier then it can not know it anymore. Hence the actual reset for `ACM_RTF ...` happens at do_finalize_rtf_i and after rtf_bold_decolour, while for everything else which needs true colour it happens later. The two reset code paths are mutually exclusive - we either reset for ACM_RTF functions or reset for the other modes, and they have different reset requirements and both need to know the original state of ATTR_BOLD.

> the role of rtf_bold_decolour in the workflow

Are you asking me to explain why it does what it does? or are you suspecting it does it differently than the original rtf code did? For the former I have a vague guess - any specific default colour which the user set for display is not necessarily what you want to copy. Therefore plain (non-inverted, and possibly bold) default colours turn to whatever the default unspecified colour of the copy is. But that's a vague guess. For the latter, I _think_ it should be identical, do you have reasons to suspect it's not?

> despite the CCL_ attributes, apply_attr_colour still tests fgi >= 256 once: is that on purpose?

Yes. This (dim) condition is highly questionable and I could not understand why it's there, therefore I left it identical (and later removed it at the dim commit, but for this commit I definitely didn't want to touch it). I do know for a fact that when it was introduced, there were no true colours - see aa821e07 .

I'll continue with the rest of the review comments shortly.

avih:
> ... there were no true colours - see aa821e07 .

Additionally, if you wonder why is it checking bold_as_colour to begin with, then you might wonder forever. The first commit which does it is not at this repository as it was merged from a branch for some 0.5 release and lost the original commit message. However, I did find and cloned a repository with the original branches and found the original commit (at branch 0.5), which, unfortunately, still doesn't say why it does it: https://github.com/avih/mintty-google-code-import/commit/ce8ac70aabb8386cf3765249edd332e47e0e2e60

avih:
> apply_bold_colour, do_rvideo for cursor and selection colours
> apply_shadow in main colour handling section

On the face of it, it should not have happened, and I might have not taken some factors into account.

But I'm not sure how many different cases you're counting here, and what they are exactly (you did give one test case, which I will test a bit later, but would still like to understand what I should be looking for as confirmation). Would you mind being super explicit about conditions and/or colours which behave differently, and, if possible, different how?

Also, importantly, do you consider them undesirable behavior which should be fixed? or is it just a note that it changed a bit but it's otherwise fine?

> I have a few ideas how to simplify the refactored code (but we may proceed here after merging an initial version, as you prefer):

In general, I don't mind doing the legwork and bring it to the state you prefer even if I disagree (at least to some degree), but if you don't mind, I'd like to understand why.

Here are my specific comments on those:

> unfold rtf_bold_decolour into winclip.c, and perhaps unify ACM_RTF_ modes

Unifying the RTF modes is possible (I think, I should first try). As for "unfolding", the reason it's where I put it is because I wanted all the attributes-to-colours code to be at one place, such that when someone makes a related change, they can see the whole picture at one place. I find it very valuable, and otherwise I don't see why it matters much where this functionality resides. But if you insist, then it can be "unfolded" back to the rtf code. Let me know your decision.

> unfold do_vbell_bg (and keep term.c: tattr.truebg = brighten(bg, fg);),
> and save ACM_VBELL_BG

It seems minor, but I put it there with a lot of intent. The reason is, like I mentioned at https://github.com/mintty/mintty/pull/676#issuecomment-326988319 (where I knew much less than today, but my feeling only became stronger since) that term_paint is an abstract function which only touches attributes. The translation of attributes to display properties like preferred user colours etc, I think, must be at the "font end" and not at the abstract core terminal.

If you noticed, until you added the bellflash bg highlight and shortly later the custom selection colours, term_paint was relatively pure with only attributes handling.

I very strongly believe it should be only about attributes, and that the bellflash bg thingy and custom selection colours (and really, selection in general) should be TATTR_* stuff which the front end renders and the core terminal does not concern itself with.

There is an issue though, because there only about 3 bits left at TATTR, and it would be one if we make attributes of bellflash and selection. It is solvable, but I didn't want to spread myself even thinner.

So for now, the least I could do is remove any notion of concrete colour manipulation for bellflash at term_paint . I would have done it for the selection colours too, but it added too much for this refactor so I left it as is, but added a comment that bellflash and selection should be attributes and rendered at the front end.

> maybe unfold do_reverse and do_blink_i (and check whether blink should set bold/colour at all),
> and save ACM_TERM

When you say "save", what are we saving exactly? Is it less code? easier to understand code? better performance? Maybe some other thing which is being "saved"?

Also, do notice that at the refactor commit message I did say that the long term goal is to reduce the number of different modes. However, the proper way to do that, IMO, would be to make all the places do the same thing. I.e. NOT by reducing the unified code to the lowest/smallest common denominator of places which otherwise do different things.

Otherwise, we'll be letting this monster grow again, and we don't want that. The goal is to have the least unique _requirements_ as possible, and the first step is putting as much as we can put at one place, then changing the _clients_ to be able to use more generic functionality (as a small example - make the rtf copy thing work with true colours too, and be done with that index-only early-exit nonsense).

Once you take out things which only one or two modes need, you start losing the big picture again, and it becomes harder to have more consistent behavior globally.

> unfold apply_bold_colour to have changes at a glance,

This is questionable IMO. I think that each non-trivial functionality which can be contained, should be contained (in a function). In this case, the goal of apply_attr_colour is to run the "sequence" of going through the different attributes and applying them. Whenever the application of an attribute is non trivial, I moved it to a function. This way you can get the overview by skimming apply_attr_colour, and more details, if you care, at apply_bold_colour, etc.

As a demonstration of how this "unfold to see at a glance" thing becomes a monster quickly, take a look at winclip before the refactor. Did you know, for instance, that `win_copy` had more than 350 lines of code, and more than 300 of them (smack in the middle) are completely skipped when copying plain text _not_ as rtf? But the rtf copy code is "unrolled" to have a better glance how rtf work, right? Well, I can tell you that, there's no glance that can cover the rtf code or the function which contains it which copies also non-rtf code.

So, whenever there was a non trivial functionality which could be described as a "unit" where the containing code does not care about its internal details, it was moved to a function.

> and maybe get the reset_bold logic clearer?

Other than my earlier explanation, if you have suggestions on how to make it clearer, I'm listening.

> implement mapping of true colour → palette for RTF export

Yes, but not now. Recall, the refactor is about _not_ changing functionality. It's true that it does make changing functionality easier, like with bold (or dim), but let's first finish what we started (i.e. conclude the refactor approach, and bold, and maybe dim, and then merge), and after that the sky is the limit.

avih:
> The refactoring patch is not completely a no-op (although the difference makes it more consistent);
> issues:
> 
>     apply_bold_colour, do_rvideo for cursor and selection colours
>     apply_shadow in main colour handling section

So, I tried to understand what you mean by this, and I could not. Again, could you please be super explicit about what has changed, or, alternatively, give me an exact procedure to run through without modifying the code, and then tell me exactly what to look at?

Nevertheless, I did try to follow your instructions:

> To demonstrate, add the following ...

> To demonstrate, add the following in wintext.c, behind the #ifdef debug_underline section:
> 
> #define debug_bold
> #ifdef debug_bold
>   if (xwidth > 1) {
>     force_manual_underline = true;
>     ul = 0x80E02020;
>   }
> #endif
> 
> Run mintty -o BoldAsColour=1 -o BoldAsFont=0
> echo -e "\e[?5h"; cat xbold
> then select the output

Other than not being able to understand what it does and why demonstrating an issue requires patching the code, I did try it and did the following:

1. I assumed you're testing a commit which includes the refactor and unify-rvideo patches but not the dim patch onward (as the dim and later commits obviously make real changes).
2. I interpreted "behind the #ifdef debug_underline section" as below it, i.e. just before the `/* Underline */` comment, yes? 
3. I tested as follows:
    - I applied the underline test patch to the original mintty sources before the `add typedef unsigned long ...` commit and built mintty.
    - I applied the `add typedef unsigned long...` commit and the refactor commit and the bugfix commit and the unify rvideo commits, and then the underline patch on top of them (i.e. without the dim and bold changes), and then built mintty.
4. I deleted ~/.mintty.rc, then increased the font size to 12 without any other config changes (it was and remained Lucida Console).
5. I ran it as you requested, and then selected the output, but I don't _think_ I'm seeing a difference, and I'm not sure what to look at exactly.

Here's what I see: left-side: orig + underline patch (selected), right-side: up to the rvideo commit and then the underline patch (selected).

xbold-under-orig-vs-under-refactor-and-rvideo

Could you please point me to where and what the difference is?

Or even better, give me a _minimal_ test procedure which doesn't involve patching the code? Thanks.

mintty:
Just some quick answers, more details tomorrow.
By unfolding I meant only the specific part. For example the decolour function is only called in one call situation, I think, and could be taken out from wintext.c. In winclip.c, the apply_attr... function would still be called, just the decolour stuff would be handled explicitly afterwards.
The purpose of the "debug" patch is to make it visually clearer when the shadow bold was applied rather than font bold.

avih:
> Could you please point me to where and what the difference is?

OK, I missed it, it's the bold/dim at the second 16M#red line, yes? where the original mintty does not display it as bold, and with the refactor it does display it bolded (well.. shadow), right? I.e. the original mintty is buggy and the refactor commit fixes it?

Also, it shows also without selecting the output.

Here's a smaller test case with the original mintty and the up-to-rvideo-commit, both without the underline test patch, ran it the same as you requested, then issued the following:
```sh
echo -e "\e[?5h"; printf "\033[1;38;2;255;0;0mbold \033[2m dim\e[m\n"
```
And here's what I see (top is the original mintty, bottom is the refactor+rvideo commits), without selecting anything:
xbold-plain-orig-vs-plain-refactor-and-rvideo simple-command

I think it should be considered a bug of the refactor (or the rvideo commit), but I'll look into that to know better. Thanks. Do let me know if you see other differences.

avih:
> The purpose of the "debug" patch is to make it visually clearer when the shadow bold was applied rather than font bold.

Is it meant to mark all the lines where you see a difference (i.e. probably a bug) with the new refactor? Because (so far) I can only see a difference only at the bold/dim columns and only at the second 16M#red line ...

avih:
> I think it should be considered a bug of the refactor (or the rvideo commit), but I'll look into that to know better

So, this is a bug (at the original mintty) which combined three different sloppy tests for default colours, all of which were fixed at the refactor by the move to the more strict and accurate CCL_ macros. Here are the relevant parts of the original code:

```c
  if (term.rvideo) {
    //   Bugs #1 and #2: the code which tests >= 256 was written before there
    //   were true colours, and assumes that >= 256 means a default colour, but
    //   in this case it's actually true colour. Nevertheless, it flips bit 2.
    if (fgi >= 256)
      fgi ^= 2;     // (BOLD_)?FG_COLOUR_I <-> (BOLD_)?BG_COLOUR_I
    if (bgi >= 256)
      bgi ^= 2;     // (BOLD_)?FG_COLOUR_I <-> (BOLD_)?BG_COLOUR_I
  }
  if (attr.attr & ATTR_BOLD && cfg.bold_as_colour) {
    if (fgi < 8) {
      if (!cfg.bold_as_font)
        apply_shadow = false;
      fgi |= 8;     // (BLACK|...|WHITE)_I -> BOLD_(BLACK|...|WHITE)_I
    }
    // ---- bug #3: another (different) sloppy test for default colours
    // should be fgi >= 256 && fgi < TRUE_COLOUR, but it happened to work as
    // long as nothing flipped bits at the value TRUE_COLOUR, but rvideo did
    // flip, therefore it was indeed both >= 256 and != TRUE_COLOUR, and yet was
    // also not a default colour as the test was trying to assert.
    else if (fgi >= 256 && fgi != TRUE_COLOUR && !cfg.bold_as_font) {
      apply_shadow = false;
      fgi |= 1;     // (FG|BG)_COLOUR_I -> BOLD_(FG|BG)_COLOUR_I
    }
  }
```

So what happens is that it started as TRUE_COLOUR, but then the original rvideo code accidentally thought that >= 256 means default colour and flipped (set) bit 2 even though it shouldn't have (but so far no bugs).

Then the test for TRUE_COLOUR at the bold code below was sloppy, and tested != TRUE COLOUR rather than < TRUE_COLOUR. Therefore it "detected" (wrongly) that it's a default colour and applied `fgi |= 1` which of course doesn't have any effect on true colours which are stored at `attr.truefg`.

Therefore the boldening (shadowing in this case) ended up false thinking that it would applied colour, but in fact didn't apply colour, and also didn't apply boldening, hence it didn't show as bold (shadow) before the refactor.

What the refactor does on this case avoids all three bugs:
The rvideo code tests explicitly that it's a default colour rather than a sloppy >=256:
```C
  if (do_rvideo_i && term.rvideo) {
    if (CCL_DEFAULT(fgi))
      fgi ^= 2;     // (BOLD_)?FG_COLOUR_I <-> (BOLD_)?BG_COLOUR_I
    if (CCL_DEFAULT(bgi))
      bgi ^= 2;     // (BOLD_)?FG_COLOUR_I <-> (BOLD_)?BG_COLOUR_I
  }
```

And the bold application code is similarly strict and correct (the return value is the equivalent of the earlier `apply_shadow`):
```C
  if (cfg.bold_as_colour) {
    if (CCL_ANSI8(*pfgi)) {
      *pfgi |= 8;     // (BLACK|...|WHITE)_I -> BOLD_(BLACK|...|WHITE)_I
      return cfg.bold_as_font;
    }
    if (CCL_DEFAULT(*pfgi) && !cfg.bold_as_font) {
      *pfgi |= 1;     // (FG|BG)_COLOUR_I -> BOLD_(FG|BG)_COLOUR_I
      return false;
    }
  }
  return true;
```

As the commit message says, it _should_ be a no-op, however the move to the CCL_ macro of course fixes bugs due to earlier sloppy tests which are all over the code, and which rear their heads on various edge cases, such as bold true colour combined with reverse video and BoldAsFont=no.

I don't consider it a bug of the refactor, even though indeed it makes it a no-op. But such changes can only be for the better.

avih:
> The refactoring patch is not completely a no-op (although the difference makes it more consistent);
> issues:
> 
>     apply_bold_colour, do_rvideo for cursor and selection colours
>     apply_shadow in main colour handling section

Like what you expect from users in bug reports, please tell me only exactly what/where the difference is, how to reproduce what you're seeing, with a screenshot if possible. I appreciate additional information and/or analysis, but not before I can reproduce it and understand what you consider an issue (at this case "issue" means it behaves/looks different than before).

We've concluded that bold + true colour + rvideo + BoldAsFont=no is a bugfix, and this also applies to custom selection colours which are using the `truecolour` function which had the same bugs.

However, I don't have an obvious explanation to the change in cursor colours. Would you mind giving me reproducible steps to see the difference?

mintty:
>> The purpose of the "debug" patch is to make it visually clearer when the shadow bold was applied rather than font bold.

> Is it meant to mark all the lines where you see a difference (i.e. probably a bug) with the new refactor?

No, it marks shadowed boldening as opposed to bold font, whether properly applied or not.

And I agree that the patch makes things better, nor worse. Although in some cases boldening (font or shadow) changes with selection, but that may be attributed to the changed colouring and might be considered OK.

Here are 4 screenshots; old/new `mintty -o BoldAsFont=0 -o BoldAsColour=1 -s 74x20`,
then old/new with selection:

old
new
oldsel
newsel

avih:
> Although in some cases boldening (font or shadow) changes with selection, but that may be attributed to the changed colouring and might be considered OK

Right. Recall, it's not about fixing all colour issues. Just unifying the code to make future fixes easier, while fixing some obviously buggy edge cases with the CCL macros. The correct way to make sure selection doesn't change boldness would be to make selection an attribute and apply like any other attribute for ACM_TERM, as well as additionally maybe reconsidering what selection with custom colours mean (it _seems_ to me that the custom selection colours commit was a rather hasty one without much thought given to how it may or may not change boldness when only bold as colour is true) . But this is not for this patch.

Specifically for the selection issues, with only BoldAsColour=yes the ANSI/Default colours only get colour (these are "plain", "ansi#1" and "256#1" and the second-to-last "bold" line), so when you apply custom selection colours, together with brighten, it becomes true colour which now gets shadow/bold font. That's why the one with the selection which has all the colours true colours also gets shadow for all the bold cases. Which, as you said, is more correct than before.

However, while I do understand why selection could be better, as `truecolour(..)` now goes through the unified code, the refactor doesn't get near the cursor colours, which is why I'd still like to be able to reproduce the case which ends up with different cursor colours, and understand why it happens. Hopefully it's yet another implicit bug fix, but I'd like to make sure it's not a real bug. So far I don't think you posted steps to reproduce the cursor colour issue.

mintty:
I have to correct my performance results a bit, sorry. The thing is that the refactored code is not called per character but per output chunk. A chunk is a sequence of consecutive characters with the same attributes (see break_run in term.c).
With a test file that changes attributes for each character (which is the worst case, however), like
`abababababababababababababababababababababababababababababababab`
where each b has a bold attribute, and e.g. 2000 such lines, I run
`time cat ...` a dozen times and sum up the "real" time reports.
On a 32-bit processor, this yields a performance penalty of ~6%. No penalty measured on a 64-bit processor.
I'm now awaiting a modified PR with the first 3 patches (optionally enhancements as suggested, or none for now) to merge.

avih:
> The thing is that the refactored code is not called per character but per output chunk

While true, I _think_ that what you're measuring is more optimistic than the actual performance (regardless of regression or not), because not only is it not called for every char (as you rightly noted), it's also not called for every chunk, as it's called only when the window is refreshed, which could be orders of magnitude less than the size of the output that the core terminal processes. This is _as far as I understand from brief analysis of the code_ .

I.e. if you dump a gigabyte of text, then the core terminal will process all of it, but the display code, i.e. `win_text` will be called on only a fraction of the output which mintty actually displays occasionally.

Which is why I think that from your 2000 lines which you dumped, a lot less went through `win_text`. I also don't know if this processing is blocking the terminal in a way which `time` would take into account.

So, depending on $whatever which needs careful analysis, the actual regression could be worse (or lesser? not sure).

Nevertheless, performance is important for me too, and I'll try to do a followup after careful examination of the performance and after trying to understand what the bottlenecks are. But indeed, this is not for this PR.

> I'm now awaiting a modified PR with the first 3 patches (optionally enhancements as suggested, or none for now)

So with the rtf true-colour bug fix, but without the unified rvideo commit?

The unified rvideo is fairly trivial to verify as there are maybe 3-4 places which call `win_get_colour` (which is where the unification is), and it's trivial to check them one by one and make sure nothing is affected negatively.

As for enhancements, unless I forgot something which I was convinced of (entirely possible considering the walls of texts above ;) ), I don't think I consider the "unfolding" an enhancement., and I think I replied to your other concerns (explanation about bold_decolour and reset_bold were already commented at the code).

So as far as I'm concerned, you can take the first 3 or 4 commits as is. I really think it should be 4 (i.e. including the unified rvideo) but up to you.

Let me know if you want the first three or four, and I'll remove the rest from this PR such that you can merge it as is.

avih:
Also, as for what you were _actually_ measuring, I think it's the speed at which the core terminal processes the lines. I.e. the `win_text` performance effect on the number you're getting is really small.

I think that the better way to measure it (and which I tried some time ago and which was only partially successful) would be to measure how quickly you can _refresh_ the _display_. I.e. remove some 15ms timer which is someplace I don't recall, let it refresh as quickly as possible while making sure the `win_text` code actually needs to do some work, and measure the throughput in units of refresh iterations per second.

mintty:
> Let me know if you want the first three or four, and I'll remove the rest from this PR such that you can merge it as is.

Yes, 4, with the rvideo but without the comment
`// should we consider term.rvideo? win_get_colour would.` in termout.c
as the colour configuration abstracts from other modes, this is also consistent with xterm.

avih:
> the colour configuration abstracts from other modes, this is also consistent with xterm

Cool. Thanks, I wasn't sure. I did change it to this, make it clear for others too:
```c
    c = i < COLOUR_NUM ? colours[i] : 0;  // should not be affected by rvideo
```
If you want to remove the comment completely, let me know and I will.

I also added missing spaces at one place -  `colours[i^2]` to `colours[i ^ 2]`.

Other than this, it's identical to before.

avih:
Thanks.

So now we have the bold and dim discussions left. Let's leave dim for last, but we will need to discuss it eventually because it currently uses BoldAsColour incorrectly, and has always been using it incorrectly, despite evolution of its code.

But if we clean up bold, then I think it should not depend on BoldAsColour anymore.

mintty:
I've repeated performance test, this time with a test file fitting roughly on the screen (~50 lines), still changing attribute for each character, output 1000 times. No performance penalty on either processor.

avih:
By the way, I noticed that you merged rather than rebase and merge. This creates non linear git history which could make it harder to track code evolution. Sometimes there's no choice, but if there is, I think it's better to rebase before merge.

I didn't keep rebasing my commits such that it's clear they're not changing, but next time you merge a PR, you could request the author to rebase, handle issues/conflicts if there are any, and then force push.

Alternatively and more simple, if there should be no conflicts, you can just choose rebase-and-merge option at the github interface while merging (it's one of the drop-down options of the merge button).

This would result in linear history which is easier to navigate, track, and evaluate.

mintty:
Linear history sounds good; I use only a minimal feature set of the git world which I consider somewhat over-engineered. But I wouldn't have expected anthing non-linear to occur as I don't remember a merge conflict to be announced when I merged. I'm not really familiar with what rebasing really means and why it could be necessary when there is no merge conflict. A pointer to a short (!) explanation would be appreciated :)

avih:
> A pointer to a short (!) explanation would be appreciated :)

Unfortunately, I don't think any explanation with git can be short. Nevertheless, IMHO it's worth spending the time to understand it better as other than having to use it as part of maintaining a project, it can actually do a lot of nice things which can help your interaction with the code.

But I'll try to make "short" in images.

Currently this is how the mintty history looks (if you open the repo with `gitk` or run `git log --graph --pretty=oneline`:

mintty-merges

As you can see, there's a divergence/split at the history at `prevent initial bogus font zooming (#708)`. This is because that's what the mintty master was when I created the refactor branch.

But since I created the branch, and started adding commits (`add typedef unsigned ...` and `refactor: ...` etc), the master branch has progressed independently from my branch with `fix restore wrong win HWND` etc.

And now we have two "threads" of history, my branch which adds refactor commits on top of `prevent initial bogus font zooming` and the master branch which adds different commits on top of the same `prevent initial bogus font zooming`.

And when you merged the branches, it well.... merges the two history threads into one by merging the code.

The problem with this is that if someone notices a bug at the merge commit, now you don't know which way to go to backwards to find its source, as it could come from either of the history threads.

However, if all the commits were rebased before the merges, then it means the commits at the branch were "moved" on top of all the latest commits of mintty master instead of progressing independently. It's rewriting the history of the branch to make it look (and actually also be) as if all the commits at the branch were added after all the commits at the master branch. I.e. like it never diverged at all.

In pictures, it would have looked like this:

mintty-rebasees

And now if you have a bug and you want to test which commit introduced it, you have a linear history which you can bisect easily.

It's also possible to bisect diverged history, but if there's no reason for the history to be diverged in the first place, and if rebase if possible (mostly it is, except for big divergence between the branch and the master branch, which has to be solved somehow anyway), then it's nicer to first rebase.

Oh, wait, I think I did find a short explanation after all. [Here](https://git-man-page-generator.lokaltog.net/) you can find extremely simple answers to any git question you might have. Really!

mintty:
> answers to any git question

I see you understand me. Thanks for the link.

So "Rebase and merge" would do just what you describe and not apply any magic on the sources?
Can the current fork still be sorted out?

avih:
> So "Rebase and merge" would do just what you describe

Yes.

> not apply any magic on the sources?
> Can the current fork still be sorted out?

I don't understand these questions.

Regardless, it is important to understand that rebasing does lose some history details forever. For instance, at the second screenshot it's impossible to know that my refactor branch was developed without the `fix restore wrong win HWND` commit at the branch history.

If for some reason they are incompatible, for instance my changes assume that some variable X has an initial value of FOO, but the `fix restore wrong win HWND` commit changes the default value of X to BAR - and there's no merge conflict because they touched different files, then it's harder to understand why my commits assumed a value of FOO while clearly they're added after the commit which changed it to BAR.

Which is why I said you should ask the author to rebase, i.e. that the author make sure their changes work correctly with the latest version of mintty.

Or you can do that too with the "rebase and merge" button if you look at the changes and decide that they don't get anywhere near anything which happened since on the main branch.

Ultimately, making sure the PR is compatible with latest mintty is a responsibility both of the PR author who wrote the code, but also yours because it's your project and you'll have to deal with the consequences.

So rebase is not a magical thing which makes everything simpler. Unless you know it can be rebased safely, the safer option would be to do a normal merge without rebase, deal with the inconvenience of non linear history, but have a more accurate history which really reflects how the code looked when the PR was authored.

mintty:
The refactoring step fails on the italic attribute. I don't see why at the moment.

mintty:
No, sorry, it wasn't this patch; forget about it.

avih:
It's 43ec70b6 .

mintty:
Thanks, this had uncovered a lurking code issue. Fixed. Will remove these comments as they don't really belong here.

avih:
While the comments which were removed indeed ended up unrelated to this PR, I think it'd be a good practice to not remove comments (especially of others, but also yours) unless they reveal some secrets or violate whatever you consider a violation.

They do serve as records of the discussions which took place, and it's easier to skip irrelevant ones than to dig from the grave comments which were removed ;)

mintty:
Released 2.8.2.

mintty:
Released 3.5.3.

PR #713, opened 2017-11-09
fix restore wrong win HWND

vgoah10:
fixed the bug when using sessions switcher, the minimum window can not restore to normal size


mintty:
This fix looks correct, but I cannot reproduce the wrong behaviour that it's supposed to fix.

mintty:
Reproduced; not sure why I didn't before.
Thanks.

PR #714, opened 2017-11-09
Pr bold demonstration

avih:
Note: this PR demonstrates the bold config improvements, while knowingly ignoring the behavior of dim, copying to rtf, combined characters, custom selection colours, and visual bell background-highlight style. PR #710 does the same as this PR while taking those into account.

Terminology:
- Bold Font: font variant thicker than the normal font. Sometimes we don't have it or can't use it.
- "Shadow" (same as "overstrike"): emulate bold font using a normal font (redraw with 1px offets to render it thicker). Typically inferior to bold font, but has the advantage that we can always use it.
- "Thick": use bold font if possible, and if not possible then use shadow.

Colour classes:
- "ANSI": the normal 8 ansi colours (0-7).
- "Default": default terminal foreground colour. Not part of the ANSI colours.
- "Other": colours 8-255 and true colours (and search/IME/etc colours too).

Current mintty behavior using the above shadow/thick terminology:
```
     Config      |       Effect on the colour classes        |
---------------- | ------------------------------------------| ----------------
AsColour  AsFont | ANSI             Default         | Other  | Notes
--------  ------ | ---------------  --------------- | ------ | ----------------
no        no     | shadow           shadow          | shadow |
no        yes    | thick            thick           | thick  |
yes       no     | colour           colour          | shadow | default config
yes       yes    | colour + thick   thick           | thick  | xterm compatible
```
Problem 1:
- We should not have "shadow" at the table as it means we force shadow even if we have a suitable bold font. This is an undesirable behavior.

This issue is also affacting the default configuration (yes/no). "other" colours are not rare these days, as many apps support 256 colours or more. Currently bold "other" defaults to shadow even if we have a suitable bold font.

However, sometimes (never?) the user might prefer shadow over bold font.

Solution part 1 (commit `bold config: add ThickenForcesShadow`):
- Turn all "shadow" at the table to "thick" as the default behavior.
- Add a config which can force all "thick" to be "shadow".

Now the behavior is (with the same shadow/thick terminology):

With ThickenForcesShadow=no (the default):
```
AsColour  AsFont | ANSI             Default         | Other  | Notes
--------  ------ | ---------------  --------------- | ------ | ----------------
no        no     | thick            thick           | thick  |
no        yes    | thick            thick           | thick  |
yes       no     | colour           colour          | thick  | default config
yes       yes    | colour + thick   thick           | thick  | xterm compatible
```
And with ThickenForcesShadow=yes :
```
AsColour  AsFont | ANSI             Default         | Other  | Notes
--------  ------ | ---------------  --------------- | ------ | ----------------
no        no     | shadow           shadow          | shadow |
no        yes    | shadow           shadow          | shadow |
yes       no     | colour           colour          | shadow |
yes       yes    | colour + shadow  shadow          | shadow |
```

(We'll assume ThickenForcesShadow=no from now on. It's clear what "yes" would do)

Problem 2:
- The first two rows at the table have identical outcomes. It's a shame to give up quarter of our possible combinations, especially when we're missing a useful one: both ANSI and default are "colour + thick":

Solution part 2:
- Use the first table row to host the missing combination:
```
AsColour  AsFont | ANSI             Default         | Other  | Notes
--------  ------ | ---------------  --------------- | ------ | ----------------
no        no     | colour + thick   colour + thick  | thick  |
no        yes    | thick            thick           | thick  |
yes       no     | colour           colour          | thick  | default config
yes       yes    | colour + thick   thick           | thick  | xterm compatible
```
Problem 3:
- Now we have 4 effective combinations, but the table is even more confusing - the effect of the config values on the outcomes seem random.

Solution part 3 (commit `bold font/colour: improve control` is 2+3):
- Swap the behavior between no/no and yes/yes:
```
AsColour  AsFont |      ANSI            Default     | Other  | Notes
--------  ------ | ---------------  --------------- | ------ | ----------------
no        no     | colour + thick            thick  | thick  | xterm compatible
--------  ------ | ---------------  --------------- | ------ | ----------------
no        yes    |          thick            thick  | thick  |
yes       no     | colour           colour          | thick  | default config
yes       yes    | colour + thick   colour + thick  | thick  |
```

That's better. Except for no/no which is "special" and behaves like xterm, the three other combinations are logical - AsFont and AsColour are independent and have a logical outcome - AsColour only affects colour, AsFont only affects thicker font rendering, and they can be combined logically.

(we're ignoring the fact that we still can't use colour for "other")

Arguably it's also less surprising to have a special no/no than a special yes/yes like we had before.

And of course, the user can still use ThickenForcesShadow=yes to turn all the "thick" at the table into "shadow", while leaving AsColour and AsFont with the same logical outcomes.

### DIFF
--- a/docs/mintty.1
+++ b/docs/mintty.1
@@ -912,21 +912,46 @@ Note that font smoothing may be affected by Windows-generated underline modes.
 
 .TP
 \fBShow bold as font\fP (BoldAsFont=no)
-When this option is enabled, the ANSI bold (or 'intense') text attribute is
-shown as a bold-style font.  Where a bold variant of the selected font that
-has the same width as the base font is available, that is used; otherwise, the
-bolding is simulated by rendering the text twice with a one-pixel offset.
+Display the ANSI bold (or 'intense') text attribute as a bold-style font. Where
+a bold variant of the selected font that has the same width as the base font is
+available, that is used; otherwise, the bolding is simulated by rendering the
+text twice with a one-pixel offset.
+
+This option is not fully independent. See \fBBold Behaviour\fP below.
 
 .TP
 \fBShow bold as colour\fP (BoldAsColour=yes)
-By default, text with the ANSI bold attribute set is displayed with a
-different colour, usually with increased brightness (mapping ANSI colours 
-0..7 to their bright variants 8..15). This can be disabled here.
+Display the bold text attribute with a different colour, usually with increased
+brightness - mapping ANSI colours 0..7 to their bright variants 8..15, and the
+default colour to the value of option \fBBoldColour\fP (or to an automatic
+bright colour if BoldColour is not set, or to the bold colour set by an OSC
+escape sequence).
 
-Note that when \fBBoldAsFont\fP is enabled, only bold text in one of the eight
-ANSI colours has its colour changed, i.e. bold text without an explicitly
-specified colour is shown with a bold font only.  This matches \fBxterm\fP behaviour.
-Note also setting \fBBoldColour\fP.
+This option is not fully independent. See \fBBold Behaviour\fP below.
+
+.TP
+\fBNote: Bold Behaviour\fP
+When the the bold text attribute is set, mintty displays it while
+distinguishing between three classes of colours:
+.br
+\(en \fBANSI8\fP: The ANSI colours 0..7 (used for SGR 30..37).
+.br
+\(en \fBDefault\fP: The default terminal foreground colour (not part of ANSI8).
+.br
+\(en \fBExtended\fP: True colours and the rest of the 256 colours palette.
+.br
+
+The colour classes are affected by the bold text attribute as follows:
+.br
+\(en Extended colours are always shown with a bold font only.
+.br
+\(en When both BoldAsFont and BoldAsColour are disabled, mintty engages a mode
+similar to the xterm default behaviour: ANSI8 is displayed with a bold font and
+a different colour, Default colour only uses a bold font.
+.br
+\(en Otherwise, the ANSI8 colours and the default colour are considered one unit,
+where BoldAsFont and BoldAsColour have independent effect on this unit - such
+that it's possible to choose only bold font, or only different colour, or both.
 
 .TP
 \fBAllow blinking\fP (AllowBlinking=no)
--- a/src/wintext.c
+++ b/src/wintext.c
@@ -443,7 +443,8 @@ win_init_fontfamily(HDC dc, int findex)
     ff->fontflag[i] = 0;
   }
 
-  ff->bold_mode = cfg.bold_as_font ? BOLD_FONT : BOLD_SHADOW;
+  // if initialized as BOLD_SHADOW then real bold is never attempted
+  ff->bold_mode = BOLD_FONT;
   ff->und_mode = UND_FONT;
   if (cfg.underl_manual || cfg.underl_colour != (colour)-1)
     ff->und_mode = UND_LINE;
@@ -1510,49 +1511,6 @@ win_text(int x, int y, wchar *text, int len, cattr attr, cattr *textattr, ushort
   if (attr.attr & ATTR_WIDE)
     char_width *= 2;
 
-  uint nfont;
-  switch (lattr) {
-    when LATTR_NORM: nfont = 0;
-    when LATTR_WIDE: nfont = FONT_WIDE;
-    otherwise:       nfont = FONT_WIDE + FONT_HIGH;
-  }
-
-  if (attr.attr & ATTR_EXPAND)
-    nfont |= FONT_WIDE;
-  else
-  if (attr.attr & ATTR_NARROW)
-    nfont |= FONT_NARROW;
-
-#ifdef debug_bold
-  wchar t[len + 1]; wcsncpy(t, text, len); t[len] = 0;
-  printf("bold_mode %d attr_bold %d <%ls>\n", ff->bold_mode, !!(attr.attr & ATTR_BOLD), t);
-#endif
-  if (ff->bold_mode == BOLD_FONT && (attr.attr & ATTR_BOLD))
-    nfont |= FONT_BOLD;
-  if (ff->und_mode == UND_FONT && (attr.attr & ATTR_UNDER))
-    nfont |= FONT_UNDERLINE;
-  if (attr.attr & ATTR_ITALIC)
-    nfont |= FONT_ITALIC;
-  if (attr.attr & ATTR_STRIKEOUT
-      && !cfg.underl_manual && cfg.underl_colour == (colour)-1)
-    nfont |= FONT_STRIKEOUT;
-  if (attr.attr & TATTR_ZOOMFULL)
-    nfont |= FONT_ZOOMFULL;
-  another_font(ff, nfont);
-
-  bool force_manual_underline = false;
-  if (!ff->fonts[nfont]) {
-    if (nfont & FONT_UNDERLINE)
-      force_manual_underline = true;
-    // Don't force manual bold, it could be bad news.
-    nfont &= ~(FONT_BOLD | FONT_UNDERLINE);
-  }
-  if ((nfont & (FONT_WIDE | FONT_NARROW)) == (FONT_WIDE | FONT_NARROW))
-    nfont &= ~(FONT_WIDE | FONT_NARROW);
-  another_font(ff, nfont);
-  if (!ff->fonts[nfont])
-    nfont = FONT_NORMAL;
-
   attr = apply_attr_colour(attr, ACM_TERM);
   colour fg = attr.truefg;
   colour bg = attr.truebg;
@@ -1599,6 +1557,49 @@ win_text(int x, int y, wchar *text, int len, cattr attr, cattr *textattr, ushort
 #endif
     }
   }
+
+  uint nfont;
+  switch (lattr) {
+    when LATTR_NORM: nfont = 0;
+    when LATTR_WIDE: nfont = FONT_WIDE;
+    otherwise:       nfont = FONT_WIDE + FONT_HIGH;
+  }
+
+  if (attr.attr & ATTR_EXPAND)
+    nfont |= FONT_WIDE;
+  else
+  if (attr.attr & ATTR_NARROW)
+    nfont |= FONT_NARROW;
+
+#ifdef debug_bold
+  wchar t[len + 1]; wcsncpy(t, text, len); t[len] = 0;
+  printf("bold_mode %d attr_bold %d <%ls>\n", ff->bold_mode, !!(attr.attr & ATTR_BOLD), t);
+#endif
+  if (ff->bold_mode == BOLD_FONT && (attr.attr & ATTR_BOLD))
+    nfont |= FONT_BOLD;
+  if (ff->und_mode == UND_FONT && (attr.attr & ATTR_UNDER))
+    nfont |= FONT_UNDERLINE;
+  if (attr.attr & ATTR_ITALIC)
+    nfont |= FONT_ITALIC;
+  if (attr.attr & ATTR_STRIKEOUT
+      && !cfg.underl_manual && cfg.underl_colour == (colour)-1)
+    nfont |= FONT_STRIKEOUT;
+  if (attr.attr & TATTR_ZOOMFULL)
+    nfont |= FONT_ZOOMFULL;
+  another_font(ff, nfont);
+
+  bool force_manual_underline = false;
+  if (!ff->fonts[nfont]) {
+    if (nfont & FONT_UNDERLINE)
+      force_manual_underline = true;
+    // Don't force manual bold, it could be bad news.
+    nfont &= ~(FONT_BOLD | FONT_UNDERLINE);
+  }
+  if ((nfont & (FONT_WIDE | FONT_NARROW)) == (FONT_WIDE | FONT_NARROW))
+    nfont &= ~(FONT_WIDE | FONT_NARROW);
+  another_font(ff, nfont);
+  if (!ff->fonts[nfont])
+    nfont = FONT_NORMAL;
 
   SelectObject(dc, ff->fonts[nfont]);
   SetTextColor(dc, fg);
--- a/src/wintext.c
+++ b/src/wintext.c
@@ -1353,21 +1353,26 @@ apply_bold_colour(colour_i *pfgi)
   // colours: ANSI (0-7), default, and other (8-256, true). We also reserve one
   // combination for xterm's default (thicken everything, ANSI is also coloured).
   // - "other" is always thickened and never gets colouring.
-  // - when bold_as_colour=no: ANSI+defaut are only thickened.
-  // - when bold_as_colour=yes:
-  //   .. and bold_as_font=no:  ANSI+default are only coloured
-  //   .. and bold_as_font=yes: ANSI+default are thickened, ANSI also coloured (xterm)
-  if (cfg.bold_as_colour) {
-    if (CCL_ANSI8(*pfgi)) {
-      *pfgi |= 8;     // (BLACK|...|WHITE)_I -> BOLD_(BLACK|...|WHITE)_I
-      return cfg.bold_as_font;
-    }
-    if (CCL_DEFAULT(*pfgi) && !cfg.bold_as_font) {
-      *pfgi |= 1;     // (FG|BG)_COLOUR_I -> BOLD_(FG|BG)_COLOUR_I
-      return false;
-    }
+  // - bold_as_font:  thicken ANSI/default colours.
+  // - bold_as_colour: colour ANSI/default colours.
+  // Exception if both false: thicken ANSI/default, colour ANSI (xterm's default).
+  bool ansi = CCL_ANSI8(*pfgi);
+  if (!ansi && !CCL_DEFAULT(*pfgi))
+    return true;  // neither ANSI nor default -> always thicken, no colouring
+
+  if (!cfg.bold_as_colour && !cfg.bold_as_font) {  // the exception: xterm-like
+    if (ansi)  // coloured
+      *pfgi |= 8;  // (BLACK|...|WHITE)_I -> BOLD_(BLACK|...|WHITE)_I
+    return true;  // both thickened
   }
-  return true;
+  // normal independent controls
+  if (cfg.bold_as_colour) {
+    if (ansi)
+      *pfgi |= 8;
+    else  // default
+      *pfgi |= 1;  // (FG|BG)_COLOUR_I -> BOLD_(FG|BG)_COLOUR_I
+  }
+  return cfg.bold_as_font;
 }
 
 // removes default colours if not reversed, returns true if still needs thickening
@@ -1376,10 +1381,10 @@ rtf_bold_decolour(cattrflags attr, colour_i * pfgi, colour_i * pbgi)
 {
   bool bold_thickens = cfg.bold_as_font;  // all colours
   // if not reverse:
-  // - if only bold_as_colour, ATTR_BOLD still thickens default fg on default bg
+  // - ATTR_BOLD always thickens default fg on default bg
   // - don't colour default fg/bg (caller interprets COLOUR_NUM as "no colour").
   if (!(attr & ATTR_REVERSE)) {
-    if (cfg.bold_as_colour && CCL_DEFAULT(*pfgi) && CCL_DEFAULT(*pbgi))
+    if (CCL_DEFAULT(*pfgi) && CCL_DEFAULT(*pbgi))
       bold_thickens = true;  // even if bold_as_font=no
     if (CCL_DEFAULT(*pfgi))
       *pfgi = COLOUR_NUM;  // no colouring


avih:
For what it's worth, I did read #468 and #478 very carefully, but unless you have specific matters which you think relate to this PR beyond what I list below, I prefer to relate to them generally. For me, few things are apparent:

- Users percieve shadow as either "bold" or "bold but less pretty". They do not percieve it as converying a different meaning than real bold font. And rightfully so, because shadow is a fallback technique which is intended to imitate bold. Not to create a new type of rendering which has its own semantics.

- Users are/were confused and unhappy with the effect of the options. Mostly bold as font seemed to refuse to respect their preference when they turn it off, as it still renders "bold" (on various cases and depending on various factors).

- It seems that your principle of "not displaying bold as non bold" is not something which everyone shares. Clearly some prefer not seing bold/shadow at all (possibly also no colour at all - who are we to dictate what people prefer subjectively?). I'm not saying mintty must allow it, but in the name of this principle (which is your presonal preference), mintty does not allow some configurations which some users would have liked. For the most part this is not really an issue anymore, as by default mintty nowdays only does colour without bold font/shadow (but 256/true are still bold/shadow which users cannot turn off).

- There were also the matters of applying bold as colour to 256/true-colours and the imprefection of the `brighten(..)` imlementation, to both of which I think I have good solutions, but it's completely off topic for this PR. So please forget I just mentioned that ;)


Back to topic. Nothing in this PR contradicts or regresses neither your principle nor anything which was achieved or fixed since. This PR does not remove boldness or shadow from any colour class under any combination of the configuration values. If anything, it does the other way - turns shadow into real bold where possible and adds different colour in addition to bold/shadow.

Additionally, it hopefully helps improving the big issue of unexpected behavior, because it makes it more logical. You can use either only bold (with automatic fallback to shadow ), or only colour, or both. simple. And when both are unchecked, that's "xterm compatibility mode". Simple to understand and document.

Therefore #468 could actually be seen as a reason to _change_ the matrix rather than reason not to change it. At least that's how I see it.

mintty:
> This issue is also affacting the default configuration (yes/no). ... Currently bold "other" defaults to shadow even if we have a suitable bold font.

Sure, because BoldAsFont is false, that's the purpose of this option.
Personally, I would agree to change the default to true, but:

> However, sometimes (never?) the user might prefer shadow over bold font.

Yes, and I'd redirect complaints to you, OK?

Your two tables for ThickenForcesShadow no/yes are not convincing, that would make the option BoldAsFont rather useless.

In your case 3 table, with its purpose of being more logical, the special case is not intuitive. How about this:
```
AsColour  AsFont |      ANSI            Default     | Other  | Notes
--------  ------ | ---------------  --------------- | ------ | ----------------
no        no     |          shadow           shadow | shadow | 
no        yes    |          thick            thick  | thick  | default config
yes       no     | colour           colour          | shadow |
yes       yes    | colour + thick   colour + thick  | thick  |
```
This is a maximum of logical structure, while maintaining the meaning of BoldAsFont. It has no xterm compatible case anymore...

We could also add a clear option (to keep it intuitive) BoldAsShadow, and reserve the current fallback ("thick") to BoldAsFont=true and BoldAsShadow=true.

mintty:
Actually "xterm compatible" is a very questionable predicate; note that xterm has an even more confusing set of config parameters:
```
	BoldAsFont ≈	allowBoldFonts
			boldMode (overridden by allowBoldFonts:false)
			↯alwaysBoldMode (overridden by allowBoldFonts:false)
	BoldAsColour ≈	boldColors : 0..7 → 8..15
	BoldColour ≈	colorBDMode : bold → colorBD
			veryBoldColors:4 : combine bold with colorBD
```
So I'd rather go with simplification (as indicated above, maybe even giving up my "principle").
Furthermore, BoldColour should be decoupled from BoldAsColour (also consistent with xterm) and perhaps even from BoldAsFont (allowing both "thick" and BoldColour, not consistent with xterm).
Then BoldAsColour, like its xterm counterpart, would only be applicable to ANSI colours 0...7. And that's fine, actually. Palette or even true colours really don't need to be tweaked, that makes no sense.
For that matter, it might be reasonable to go one step further and make a distinction between the 8/16 ANSI colours as set with CSI 30... CSI 90... for which BoldAsColour would be applicable and the corresponding palette colours 0...15 as set with CSI 38;5; / CSI 48;5; for which BoldAsColour would be ignored like for all other colours. What do you think?

avih:
> Sure, because BoldAsFont is false, that's the purpose of this option.

I didn't claim I don't know why it does what it does, just that I think it should do a slightly different thing ;)

As I tried to communicate earlier, I don't think that using shadow where a real bold font can be used serves a useful purpose or have value.

As far as I can tell and assess, users (and everyone really) perceive shadow as "bold, but less pretty". I.e. it's still bold. The fact that the program sometimes uses shadow instead of bold because it can't do real bold is just implementation details (and a commendable and fruitful effort), but IMO shadow should not be used for anything other than a fallback technique to render bold font where a real bold font cannot be used due to technical reasons.

Therefore, my first suggestion here to change the "purpose" of the "bold as font" checkbox, such that when it's checked it would mean "use bold font (and if not technically possible then use shadow)", and when it's unchecked it means "use normal weight font".

As an anecdote, do notice that the "when it's checked" part of this definition is identical to what the mintty manual says already. However, what the manual conveniently fails to mention is what happens when it's unchecked.

I assert that what users expect as the opposite of "Use bold font (or fake bold font for technical reasons)" is "don't use bold font, i.e. use normal weight", while the current "purpose" as you put it is to mean "use fake bold" - which I think is highly unexpected, not useful, undocumented, and probably the main cause for filing issues such as #468 - where users say it's a bug and you reply that it's as designed.

But it did do two things: both the expected behavior when BoldAsColour=yes where it switches ANSI between bold (or shadow) and normal weight, and that "other thing" where it forces shadow when it's unchecked even when a real bold font is available.

The goal of this commit is to separate those two unrelated effects. Bold as font would be what people expect it to be: "bold" or normal weight. and ThickenForcesShadow controls what "bold" means - which only has an effect when we render "bold".

> Personally, I would agree to change the default to true, but:

I did not suggest to change it to true. I suggested to slightly modify what it means. I think the default where it's "no" is fine.

> Yes, and I'd redirect complaints to you, OK?

I don't get the sarcasm. What would they complain about? nothing was taken away. Even if for some weird reason someone actually does prefer shadow over real bold (which I honestly think no one would want. seriously, where is shadow better than real bold?), they can still do that with ThickenForcesShadow=yes .

> that would make the option BoldAsFont rather useless.

I guess you mean that it would make the change rather useless (because if you claim BoldAsFont itself is useless at those tables, then it's clearly impossible to use BoldAsColour alone to select between three possible outcomes - so it is being used).

As for the change, this is almost true, because the change would make the difference at the default configuration (no/yes) that the bold text attributes for "other" colours will now use real bold if possible rather than always using shadow. This is in line with my general assertion that shadow is just an ugly version of bold, and that it's still bold. So if it does bold, do the best bold you can. Best bold is "thick" as defined at the first post. It could be shadow, but only if real bold is not possible.

> Your two tables for ThickenForcesShadow no/yes are not convincing, 

Other than clearly being an intermediate step towards the final table, the main reason I took the effort of writing these two tables is to be crystal clear about what it does. This is because apparently I failed to explain previously what it does, judging by your earlier questions (from #710):

> Would your "don't force shadow" policy allow to effectively disable bold mode?

and

> The BoldThickenForcesShadow setting would make the current fallback behaviour questionable which is however needed because some fonts do not support BoldAsFont .

Where these tables, together with the definition of "thick" at the first post should answer those questions hopefully clearly.

> In your case 3 table, with its purpose of being more logical

Just to make sure, are you referring to the table where ThickenForcesShadow=no ? (that's the third by my count). If yes, then I don't mind, because it is an intermediate step and the first table line is modified again later anyway.

But if you meant the last table, then I think it takes away some usefulness.

Also the main goal is always to make it more useful to users. Being logical does help, but IMO, not if you sacrifice usefulness on the altar of logic.

The last table of the first post is the best I could do to represents the most useful set of outcomes, arranged by the least surprising way.

avih:
> Actually "xterm compatible" is a very questionable predicate; note that xterm has an even more confusing set of config parameters:

While true, it does represent faithfully how xterm behaves by default (and also some other terminals only one checkbox away from the default). For compatibility sake with other terminals, I do think mintty should support this behaviour even if it appears to be a weird combination of effects. The mintty manual also mentions it as an exception for compatibility reasons with xterm (for yes/yes), and I think mintty was right to have it as a possible behavior in the first place.

The question is therefore, IMHO, "which combination of config values should result in this behavior", and I still think the no/no combination is the most suitable for it, because:
- As long as you maintain your principle of "not displaying bold as non bold" then this combination can't behave logically anyway according to the new table logic - you want bold text attribute to look different than non bold, but unchecking both options suggests it gets neither boldness nor colour (and not shadow either! ;) ).
- Even without this principle, the three other combinations are, I think, more useful for users than what no/no would imply. I.e. I do think that most users would like either bold or colour or both. So putting the xterm preset at no/no is where it hurts users the least (in addition to its value as a compatibility mode).

> BoldColour should be decoupled from BoldAsColour

I don't think they're coupled to begin with. BoldAsColour, when applied to the default colour, is technically the value of colours[BOLD_FG_COLOUR_I], and this value is set independently of when/how BoldAsColour uses it.

The colour itself is set by the BoldColour or by an OSC sequence or synthesized automatically with `brighten` on init, and used by BoldAsColour completely independently of how it's set.

> and perhaps even from .. and ... and ...

OK, I like the enthusiasm (really, I do :) ) but this is becoming a lot of changes. I'm certain many of them would be useful, but how about starting small and finishing with this PR (which really, technically is a very small change to the code which still has a meaningful and useful impact on mintty's usability IMO), and once it's in, let it sink a bit, maybe get some feedback from users, fix bugs, etc.

Once we know we're good and users don't complain, then keep taking baby steps with further improvements while trying to keep everything consistent and useful?

avih:
Corrections:


> As for the change, this is almost true, because the change would make the difference at the default configuration (no/yes) ...

Should be:
> ... (yes/no) ...

As for:
> Just to make sure, are you referring to the table where ThickenForcesShadow=no ? (that's the third by my count)

First, I meant when ThickenForcesShadow=yes, and second, I later realized you refer to the last table ("case 3" rather than "3rd table") - to which I reaplied above.

Issue/solution 2 is about improving the usefulllness, i.e. allowing more users to get the outcome they want.

Issue/solution 3 is about making it easier to get what they want - make the mapping between configuration to outcome more logical.

avih:
>> This issue is also affacting the default configuration (yes/no). ... Currently bold "other" defaults to shadow even if we have a suitable bold font.
> 
> Sure, because BoldAsFont is false, that's the purpose of this option.

So, I did some detective work, which was nasty because I couldn't compile so early versions so I had to track the code. For your convenience, here's the evolution of the bold configuration from the very initial commit of mintty. Do notice the nothing(bug) at all the tables except the last:
```
Legends:
A - ANSI, D - Default, O - Other


from 2008 (initial commit) 0077106b until just before 2011-01:
- only bold_as_bright (changed to bold_as_colour at some stage):
    - no:  A:thick         D:thick          O:thick
    - yes: A:colour        D:colour         O:nothing(bug)    (default)

2011-01-16 aa821e07 : ansi is always [also] colour, for compat with other terminals.
- changes dim blend from (!as_colour) to (!as_colour && >=256). not sure why.
- only bold_as_colour:
    - no:  A:colour+thick  D:thick          O:thick           (xterm compat)
    - yes: A:colour        D:colour         O:nothing(bug)    (default)

2011-01-16 96f6c71b : bold_as_font replaces bold_as_colour (same default behavior)
- first time "Show bold as font" is at the GUI, replaces "Show bold as colour".
- bold_mode value BOLD_COLOURS changes to BOLD_NONE
- only bold_as_font (no change other than the inverted value):
    - no:  A:colour        D:colour         O:nothing(bug)    (default)
    - yes: A:colour+thick  D:thick          O:thick           (xterm compat)

2011-01-17 964b3097 : bring back bold_as_colour, but also keep bold_as_font
- Great, now we can do font or colour or both! or even nothing! whooo!
- bold_as_colour default=yes, bold_as_font default=!bold_as_colour
- changes dim blend from (as_font && >=256) to (!as_colour || >= 256). not sure why.
- asColour asFont   ANSI          Default       Other
  no       no       nothing       nothing       nothing        (as-intended? who-cares?)
  no       yes      thick         thick         thick
  yes      no       colour        colour        nothing(bug)   (default)
  yes      yes      colour+thick  thick         thick          (xterm compat)

2015-07-01 44e83137 "fix bold display by overstriking if BoldAsFont unset"
- asColour asFont   ANSI          Default       Other
  no       no       -shadow-      -shadow-      -shadow-
  no       yes      thick         thick         thick
  yes      no       colour        colour        -shadow-       (default)
  yes      yes      colour+thick  thick         thick          (xterm compat)
```

So what do we learn?

There was a bug that other colours don't get any boldness when BoldAsColour=yes. I bet this bug was introduced together with the support for 256 colours and no one noticed it (the initial commit already supports 256 colours and already has this bug).

It is abundantly clear to me that if this bug was fixed, it would have been either automatic colour or "thick" (i.e. bold font if possible, else shadow).

We also learn that until the last table, shadow was never a purpose when there was just one config value and also not a purpose when there were two config values. BoldAsFont in its various incarnations (until before the last table) always meant and was "bold font if possible, and if not possible then shadow".

It's highly likely that when the two configs were introduced, the first line where bold doesn't have any effect belonged either to the bonus category or the "don't care" category. Because the main goal clearly was to support 3 outcomes: font/colour/both. But since we now have 2 bits and 4 possible outcomes.. just let it be nothing.

Arguably the last "fix" doesn't actually fix anything because it was not broken. Well.. except "other" when only AsColour=yes. Impossible to tell because the commit message only says "fix bold display by overstriking if BoldAsFont unset", but my guess is that it mostly intended to "fix" the no/no line, but also happened to "fix" the "other" colours to shadow when only asColour=yes .

My guess is that this is roughly when your principle of "don't display bold as non bold" was born.

My guess is also that the change from "nothing" was to "shadow" rather than "thick" probably either because it was the easiest fix (it was a one line fix), or because we didn't have forced shadow at the table before, so why not allow people who prefer shadow have it their way (really.. who are those people? seriously).

And now it's a "purpose".

Can we please make it useful for users by merging my PR?

mintty:
I'm still in the process of verifying your refactoring patch. I'm through with term.c by brute-force expanding it, but honestly, it's not straight-forward due to the many mode-dependent variants of the folded functions. I'd actually prefer a more moderate folding approach, e.g. for term.c (the bell case), most options are actually not applied, so mildly unfold some details again (for example term.c: tattr.truebg = brighten(bg, fg);). The resulting code might also be better maintainable for future considerations. (Remember I had asked for some time, and I didn't mean some hours...)
Apart from that, your investigation is amazing, including the historic overview. For the records, I've been maintaining mintty since summer 2015, my first release being 2.0.1.

avih:
> brute-force expanding it

This would be hard. Would you mind if we take the refactor discussion to #710? I'll answer the rest there.

avih:
By the way, if we don't want to surprise users with an "exception" behavior, we could always change the GUI to have 4 radio buttons instead of two checkboxes:

Show bold as:
- colour
- font
- Both
- xterm

These are, after all, exactly the things we want to give users to choose from. No weird mapping, and we only need to mention that "other" colours are always and only font. Maybe even just copy-paste the last table from the first post, while combining the AsColour/AsBold columns into one config.

avih:
> (avih)  more useful for most users by merging my PR?

> (mintty) Remember I had asked for some time, and I didn't mean some hours...

@mintty FWIW, by the quote above I meant to conclude the discussion about the bold behavior, which, in my opinion is the most important thing, and the thing I think we should do first.

Once we know how we want it to behave, the rest are just technical details.

The refactor is just technical details at this case, even it would be a nice thing to have in general also regardless of the bold behavior changes.

So did we conclude that the set of outcomes at the last table of the first post is where we want to get?
If not, I'd appreciate if you could say why do you think it would be more useful for users to have the first table rather than the last.

mintty:
No, I had kind of suggested (https://github.com/mintty/mintty/pull/710#issuecomment-343769008) to clarify the refactoring first. The reason is that the amount of non-trivial code changes of this patch is a bit bothering me and I'd like to get done with it first, and not have to analyze a variant of it again later...

mintty:
Actually, I like the new bold scheme (tested with patches 7&8 of #710);
I'm still worried that others might not like it, but we can take the risk. Comments:
* Note, that 0/0 is not really xterm mode but only xterm default mode.
* With the exception of palette colour 1 mapped to palette 9 which xterm doesn't; I guess I'll fix that now or later (needs some more tweaking of colour numbers, or another cattrflags bit).
* I do not like the Force option. How about an option called "BoldAsShadow"; it would enable shadow either as a fallback or as the thickening feature if BoldAsFont=false (default true).
* I would consider changing the default of BoldAsFont to true.
* Speaking about dim, it should not affect the background, except in inverse character attribute mode; in this case, it is disputable which of inverse and dim is applied first, so let's stay like xterm here. Blending against the default background is not intuitive for me, I'd rather blend against the current character cell background.

avih:
> Note, that 0/0 is not really xterm mode but only xterm default mode.

Indeed. It's a code name for the default xterm behavior. To emulate xterm you will need nothing less than embedding it entirely as part of the mintty code ;)

> With the exception of palette colour 1 mapped to palette 9 which xterm doesn't

Huh... that sounds super weird. It certainly looks to me like it does. See [here](https://github.com/mintty/mintty/pull/710#issuecomment-342649436) at the top-left terminal - that's xterm without any configuration values. Palette 1 is red, and it definitely gets a different colour beyond bold (easier to notice at the lower reverse output at the same terminal - how the two 31/red lines have different bg colour). Regardless, I'll take a closer look at it later and report back my findings.

> I do not like the Force option. How about an option called "BoldAsShadow"; it would enable shadow either as a fallback or as the thickening feature if BoldAsFont=false (default true).

OK, so I think there are 3 subjects here, and I'm not sure I follow you:

> I do not like the Force option. How about ...

I don't mind the name, though I'd prefer it to be semantically accurate, and I do agree that this name is awkward. But I don't think it can integrate very differently with this system, other than inverting it (e.g. by making it AllowRealBoldFont or some such), but personally I usually prefer defaults to not indicate a concrete intent, hence I chose its polarity such that it defaults to false. But the polarity is minor and I wouldn't mind either way.

As far as I'm concerned, I'd remove this option and just make it a permanent "no" like it was before the commit which started forcing shadow for some configuration values. However, I've learned to appreciate your sensitivities on this matter, so I won't suggest it. But feel free to take this suggestion anyway ;)

> it would enable shadow either as a fallback

It cannot control the fallback behavior. Nothing can. Fallback to shadow is mandatory if we want thickening and we can't use real bold font. The choices for thickening are between:
- Trying real bold, and if we can't then falling back to shadow.
- Not trying real bold and instead just use shadow unconditionally.

And currently ThickenForcesShadow=yes is the latter (which I assess the vast majority of users will find useless, and I haven't heard a single attempt to suggest otherwise), while "no" is the former. I don't think there could be other choices. Are there? Or maybe I misunderstood you?

> if BoldAsFont=false (default true)
> I would consider changing the default of BoldAsFont to true.

Are you suggesting to change the default from BoldAsColour=yes and BoldAsFont=no to both yes? I don't mind personally, other than it might surprise some users which didn't modify these values - as it would now also bolden ATTR_BOLD for ANSI/default, which it didn't do before.

But as long as they can easily return to their previous behavior, this or any other combination which doesn't force shadow would be fine as well IMO, especially if you think it would be an improvement.

> Dim ... it should not affect the background, except in inverse character attribute mode

I can live with that. Everyone can, if they ever get to see an application which uses dim in their lifetime ;)

> let's stay like xterm here. Blending against the default background is not intuitive for me, I'd rather blend against the current character cell background

I _think_ this might be a contradictory statement, as it _looks_ to me that xterm exactly blends against the default bg and not against the current cell bg. It only does so for the fg colour, but nevertheless, it looks to me that it's against the default bg. I will examine it closer and make it do what xterm does - which manages to avoid the bug which dim currently has for the default fg colour in mintty.

However, if you don't mind, I won't get to this until we're both satisfied with the bold commit[s].

avih:
Let's try again

> Note, that 0/0 is not really xterm mode but only xterm default mode.

Correct.

> With the exception of palette colour 1 mapped to palette 9 which xterm doesn't; I guess I'll fix that now or later (needs some more tweaking of colour numbers, or another cattrflags bit).

I might have misunderstood you. Did you mean that the colours 0-7 of the 256 palette don't get colour in xterm? In this case you're right. It indeed only applies colour to the "normal" fg colours 30-37 and not to their equivalent at the 256 palette, e.g. `1;38;5;3` will indeed have the same colour as without bold.

However, this is what was and what still is the mintty behavior in this "similar to xterm default" mode. The default colour and 30-37 are by far the most common use cases for bold, and mintty never tried to behave like xterm beyond them.

Personally I don't think making it the same as xterm in this regard is a goal worth chasing, because it's not specified anywhere, the palette 0-15 of the 256 palette are there for compatibility reasons only and not as real specified colours. If anything, I'd argue that xterm gets this one wrong, but it doesn't matter because no one does `1;38;5;3`. They would instead do `1;33`.

So it's a questionable goal and IMO an effort definitely not worth chasing. You gain nothing useful and add complexity to the code. Don't go there.

> I do not like the Force option.

Please be more specific: You don't like the option? I.e. what it does? or don't like the name?

> How about an option called "BoldAsShadow";

I agree it sounds nicer, but this is a bad name.

When it's set as BoldAsShadow=no then you cannot guarantee that it will never be shadow, because when you want to thicken and we can't use real bold font, then you _must_ use shadow.

It's a bad you cannot always respect this option.

However, as awkward as it might sound, with a name like BoldThickenForcesShadow, you can _guarantee_ to behave like the value suggests. Because you can guarantee to force shadow - nothing stops us from doing that, and you can also guarantee to not force shadow - that's what we already do anyway - try real bold and fallback to shadow - not forcing it.

The important thing is to choose between the _only_ two ways we can _behave_:
- try real bold and fallback to shadow.
- Use shadow without first trying real bold font.

That's it, there's no other approach to thicken text. We can't force real bold because we cannot guarantee real bold. There's no point in trying shadow and falling back to real bold because shadow will always succeed anyway.

Now, unless you think there could be other approaches to thicken text, then these are our only two options.

Pick whatever config name you want to control how mintty chooses between these two options.

If you really prefer BoldAsShadow, then I can live with this. But it cannot do anything other than choosing between "try real bold and fallback to shadow" and "use shadow without trying real bold" because these are really the only two things we can choose between.

> it would enable shadow either as a fallback or as the thickening feature if BoldAsFont=false

This I don't understand and I don't want to guess. Please be explicit and draw two tables like I did at my first post.

avih:
> it would enable shadow either as a fallback or as the thickening feature if BoldAsFont=false

And I'd really appreciate if at least this time, you'll also say _why_ you think it's better than what this PR does.

I've given plenty of reasoning why I think my suggestion is better than before, and you've given exactly zero reasons for any of your suggestions.

mintty:
> > With the exception of palette colour 1 mapped to palette 9 which xterm doesn't

(see your next comment) By palette colour in this distinct context I mean a colour set by the CSI/SGR seqeuence 38;5;0..7 as opposed to SGR 30...37
(see my test file rows ansi#1 vs. 256#1; your screenshot displays only “ANSI” colour settings). Arguably, ANSI colours 0...15 and palette colours 0...15 could be considered the same, but apparently xterm makes a difference when it comes to modifications by further attributes.

> > I do not like the Force option. How about ...

>As far as I'm concerned, I'd remove this option and just make it a permanent "no" like it was before the commit which started forcing shadow for some configuration values. However, I've learned to appreciate your sensitivities on this matter, so I won't suggest it. But feel free to take this suggestion anyway ;)

Well, as I said previously, I am willing to reduce my “principle” to a moderate level if we agree on a scheme that is both logical and simple.
I could even accept to assign “xterm default” to “0/0” even though it's neither logical nor simple.

> > it would enable shadow either as a fallback

>It cannot control the fallback behavior. Nothing can. Fallback to shadow is mandatory if we want thickening and we can't use real bold font.

Maybe I didn’t consider that thoroughly enough before (because I wanted to get rid of the burden of refactoring code masses first...), but I think my main problem is this: You suggest a paradigm of BoldAsThicken as the main issue; 
I would even agree on that but I don’t see how that could be made compatible with the current name (and associated user expectation) of BoldAsFont. I do not think a re-interpretation of Font as Thicken would be proper.

>The choices for thickening are between:

> *  Trying real bold, and if we can't then falling back to shadow.
> *  Not trying real bold and instead just use shadow unconditionally.

>And currently ThickenForcesShadow=yes is the latter (which I assess the vast majority of users will find useless, and I haven't heard a single attempt to suggest otherwise), while "no" is the former. If we'd use AllowRealBoldFont instead then yes would be the former. But other than using different names or polarity, I don't think there could be other choices. Are there? Or maybe I misunderstood you?

Ignoring AsColour for a moment, I think the difference is basically this:
your suggestion:
```
AsFont	ForceSh	|
no	no	| font/shadow
no	yes	| shadow
yes	no	| font/shadow
yes	yes	| shadow
```
my suggestion:
```
AsFont	AsShadow|
no	no	| none
no	yes	| shadow
yes	no	| font(or none)
yes	yes	| font/shadow
```
Comments to these schemes:
* If AsFont=false, you would turn current behaviour (shadow) into thick (= font/shadow), meaning you add the case of superseding shadow with font if “BoldAsFont”=false. This is counterintuitive.
* If we consider the second option roughly as “EnableShadow” in both cases, your version would give it precedence (if both are true), while mine would give BoldAsFont precedence. As we both agree that shadow should only be a fallback case, I think it should not get a precedence option.
* My scheme would allow finer control of the bold methods.

About colour, an additional option could select whether to combine colour with thick or use colour only where applicable.

> > I would consider changing the default of BoldAsFont to true.

>Are you suggesting to change the default from BoldAsColour=yes and BoldAsFont=no to both yes? I don't mind personally, other than it might surprise some users which didn't modify these values - as it would now also bolden ATTR_BOLD for ANSI/default, which it didn't do before.

BoldAsColour could by default mean BoldAsColourOnlyWhereApplicable, so we’re back at previous behaviour.
Also, existing users – if they’ve ever saved a config − are likely to have BoldAsFont=false in their config file, so they would not notice :/

> > Dim ... it should not affect the background, except in inverse character attribute mode let's stay like xterm here. Blending against the default background is not intuitive for me, I'd rather blend against the current character cell background

>I think this might be a contradictory statement, 

I meant like xterm with respect to not modifying the background (except for inverse), not about how to modify...

>as it looks to me that xterm exactly blends against the default bg and not against the current cell bg.

I’ll look too, but I think we don’t need to mimic xterm in every detail, you objected against the blending against default background before (because it may make colour actually look brighter) and I agree on that. I think blending against cell background is the nicer look.

>It only does so for the fg colour, but nevertheless, it looks to me that it's against the default bg. I will examine it closer and make it do what xterm does - which manages to avoid the bug which dim currently has for the default fg colour in mintty.

Which bug, for that matter? What I’ve just said above? So let’s drop the selective mode and do it uniformly.

>Let's try again

which makes my response take even longer...:) I’d like to work full-time on open source, but I can’t.

> > [palette colour 1 mapped to palette 9 ...]

>... mintty never tried to behave like xterm beyond them.

No, I just noticed it on this occasion and I somehow like the idea that if you use BoldAsColour, an application still has the choice to set a colour 0...15 which is not modified when boldened. But this is another change and does not have to be in scope of this issue.

>Personally I don't think making it the same as xterm in this regard is a goal worth chasing, because it's not specified anywhere, → Also the actual values of colours 0...15 are not specified anywhere...

>If anything, I'd argue that xterm gets this one wrong, but it doesn't matter because no one does 1;38;5;3. They would instead do 1;33.

If no one does this, there isn’t much to get wrong, and I think I’ll do it just for the fun of enabling the choice...

> > How about an option called "BoldAsShadow";

>When it's set as BoldAsShadow=no then you cannot guarantee that it will never be shadow, because when you want to thicken and we can't use real bold font, then you must use shadow.
>It's a bad you cannot always respect this option.

I would, because as explained above, I’d drop the restriction and allow “none” in this case, unless you take this up and depend my “principle”.

>BoldThickenForcesShadow...

Let’s agree on the scheme first and then find suitable names.

>Now, unless you think there could be other approaches to thicken text, then these are our only two options.

Well, loosening the thicken requirement as contemplated above...

avih:
> ignoring AsColour for a moment, I think the difference is basically this:
> your suggestion:

I don't know, and I'm starting to give up. Clearly we both want the same: improve mintty, but we're completely failing to communicate...

> Well, as I said previously, I am willing to reduce my “principle” to a moderate level 

Let's start from scratch and see below. Bottom line, I don't want you to compromise anything, OK? Just decide what you want and then spell it out for me so that I can abide by it. Compromises come after we both understand what the other want, but so far I'm not convinced at all that you need to compromise.

> Let’s agree on the scheme first and then find suitable names.

To this I totally agree. So let's ignore all option names and focus only on what users can do, regardless of how they can do it.

Do correct me if you think it's incorrect (really, I insist), but right now, these are the outcomes of the bold text attribute which users can choose from (each a row is an outcome, and the columns are how different colour classes are affected at each outcome).

Table "Before":
```
 ------------------------------------------| ----------------
 ANSI             Default         | Other  | Notes
 ---------------  --------------- | ------ | ----------------
 colour + thick   thick           | thick  | "xterm mode"
 -------------------------------- | ------ | ----------------
 thick            thick           | thick  |
 colour           colour          | shadow |
 shadow           shadow          | shadow |
```

I'm saying the following:
1. You have a principle you'd like to keep, and I don't want you to compromise it.
2. You might have other things you'd like to keep, and which I also don't want you to compromise.
3. I don't think shadow is useful for users if real bold can be used instead. I.e. in the terminology of the first post: IMO "thick" is always better than "shadow".
4. I think it's important to keep an outcome which is similar to xterm-default.
5. I think it's important to keep offering users combinations of thickness and colour.
6. There is an outcome I think is useful which currently is not offered:
```
      ANSI             Default    | Other
 ---------------  --------------- | ------
 colour + thick   colour + thick  | thick
```

For reasons 3-6 (i.e. except your principle and possibly other things you wanna keep), I think this would be a better set of outcomes to ofer users to choose from, both in comparison to what they can do before the PR but also regardless (note that the first two outcomes are identical to before, and the third outcome only differs at the "other" column):

```
 ------------------------------------------| ----------------
      ANSI             Default    | Other  | Notes
 ---------------  --------------- | ------ | ----------------
 colour + thick            thick  | thick  | "xterm mode"
 -------------------------------- | ------ | ----------------
          thick            thick  | thick  |
 colour           colour          | thick  |
 colour + thick   colour + thick  | thick  |
```

This would have been enough _for me_ as a good and better set of outcomes to offer to users, but I still have your principal to adhere to, and maybe other things too.

As for the principle, the only thing I know about it is how you described it yourself: "don't display bold as non-bold", which I interpret as "The bold text attribute should make it look different than if it didn't have the bold text attribute, regardless of other attributes or colours".

Now, my interpretation is knowingly a guess, and will forever be a guess until you spell it out clearly what it means in practice or otherwise. I'm literally begging you to spell it out for me explicitly.

However, judging by what mintty does currently, it seems correct. Currently (before the PR) the bold text attribute indeed always changes the look in one of four ways:
- Change its colour ("colour").
- Make it use shadow ("shadow").
- Make it real bold font and fallback to shadow ("thick").
- Change its colour and make it real bold font or fallback to shadow ("colour + thick").

So the principle that the bold text attribute changes how the text looks was served perfectly well before the PR.

And the set of outcomes which I'm suggesting also don't violate this interpretation. It only uses three ways to make it look different ("colour", "thick", "colour + thick"), but the principle as far as the interpretation go is still kept perfectly intact. Not only is it intact, it's even served in a stronger way, because all the "shadow" at the previous outcomes only become even more different: they become real bold font or bold font and additional colour.

As for other things you want to keep, if there are such, I don't think I've seen you mention them in this or other threads. And I can't read people's minds, so unless you spell those out too, I will never know what they are.

So until I know what they are, I'm assuming they don't exist.

And that's it. I think these four outcomes I'm suggesting - where two were already offered before and two are different -  are a perfectly good enough set for mintty to offer, and as far as I can tell your principle is served with utmost respect, and other than your principle you haven't mentioned anything else you want to keep.

Now it's your turn to tell me what I got wrong. I clearly got something wrong because you keep suggesting different things despite my best attempts to not make you compromise anything I know you prefer not to compromise, and you're (still) not explaining WHY you keep suggesting different things.

So what did I miss?

mintty:
I could not yet analyze your latest comment completely, but
* I confirm your interpretation of my "principle", adding "regardless of ... and config settings".
* I agree with "A, B, E, F" to be supported.

I am having a problem, though, with "A", applying colour although BoldAsColour as false.
I will think about suitable configuration to enable those modes (and perhaps others).
I also agree that shadow does not need to be explicitly selectable, so the Force option should be dropped. On the other hand, it might be a good idea to have a compatibility option OldBold.

avih:
> I am having a problem, though, with "A", applying colour although BoldAsColour is false
> On the other hand, it might be a good idea to have a compatibility option OldBold.

Wait. We're not there yet. We're at the stage of agreeing what's the set of outcomes we want to offer users to choose from. Don't spare anything, if you want some outcomes to remain for backward compatibility, just say so. We'll make every thing right eventually, but one step at a time to keep us in sync.

So is it A/B/E/F ? or is it A/B/C/D/E/F ? or is it anything else?

Let's conclude this first, and once we agreed, then move to how users can configure mintty to provide each outcome.

avih:
> I confirm your interpretation of my "principle", adding "regardless of ... and config settings".

I don't think I understand how your addition combines with my interpretation. I'd appreciate if you state it fully from start to finish in one sentence.

mintty:
I’ve checked your proposal against xterm which has 6 configuration resources.
I’ll ignore boldMode and alwaysBoldMode because they only configure the thick mode, i.e. the choice between boldfont and shadow mode.

So the remaining 4 may be relevant for us:
```
allowBoldFonts	triggers bold, except for default (colorBD) without veryBold
boldColors	impact only ANSI 0..7, no impact on bold display (thick)
colorBDMode	enables default to be coloured (if a colour is defined)
veryBoldColors	makes colorBD bold only if allow && colorBD
```
Then your table corresponds to the following xterm resource combinations:
```
        ANSI             Default    | Other  | xterm
   ---------------  --------------- | ------ | ----------------
A  colour + thick            thick  | thick  | allowBF, boldColors, ¬colorBDMode
B           thick            thick  | thick  | allowBF, ¬boldColors, ¬colorBDMode
E  colour           colour          | thick  | ¬allowBF, boldColors, colorBDMode
F  colour + thick   colour + thick  | thick  | all 4 true
```
I’ve opened xterm windows with all 16 combinations on a screen and checked which of them can be configured with your scheme. I concluded that none of the others needs to be configurable additionally.

---

Actually, I just notice that I’ve confused the nature of “colour” somehow; 
while in xterm, “Default, colour” means colorBDMode with colorBD (which is BoldColour in mintty), mintty has two “colour” methods, including “brighten” which xterm does not have. Furthermore, mintty colorBDMode is coupled to BoldAsColour which we should perhaps decouple, to be also clearer and consistent with xterm.

Maybe we’ll come to a conclusion tomorrow; maybe after the weekend during which I’m busy, sorry.

avih:
> Then your table corresponds to the following xterm resource combinations:

This is definitely good info, but I don't think I understand why we need it.

Is it to confirm or argue against offering A/B/E/F ?
Is it to find more useful outcomes to offer?
Is it to check how close our "xterm mode" to actual xterm?

And what's your conclusion from this info?

mintty:
> This is definitely good info, but I don't think I understand why we need it.

This was only a step to get a better overview (also of potential other modes to support) on the way to a conclusion, and to let you know that I'm making progress:)

OK, let's make it so. Meaning as your latest proposal, 4 modes, with obvious mapping to the existing 2 main configuration settings BoldAsFont and BoldAsColour. If people complain, we can add BoldAsShadow later as I had suggested, but not initially. Descriptions could be like this:

BoldAsFont (default... yet to be decided, let's stay false for now).
Setting this true selects the preferred display of the bold text attribute to use a bold font; if a suitable bold font is not available, an overstriking shadow mode is used instead.
If both BoldAsFont and BoldAsColour are true, both display methods are combined where applicable.
If both are false, xterm default behaviour is applied.

BoldAsColour (default true)
Setting this true selects the preferred display of the bold text attribute to use a different colour as follows: ANSI colours 0...7 are mapped to their "bright" versions 8...15, default foreground is "brightened" algorithmically. Where coloured bold display is not applicable, boldened character display is used instead (using bold font or shadow mode as described for BoldAsFont).
If both BoldAsColour and BoldAsFont are true, both display methods are combined where applicable.
If both are false, xterm default behaviour is applied.

avih:
Cool, so, I can totally live with what you suggest. So we have a base line. I'll update the commits at this PR to: 1. Remove the shadow thingy. 2. set the mapping to A/B/E/F in what you called "the obvious mapping" which I'm guessing is the last table of the first post, yes?

However, now that we're done with the set of outcomes and can focus on the configuration mapping, the mapping is actually not very obvious to me. I completely share your concern that the no/no combination is not intuitive at all.

I had two ideas for alternative control scheme, and we can easily do it in this PR or in a followup, or never:

1. Use radio buttons for our options: "Bold as Colour", "Bold as Font", "Bold as Colour and Font", "xterm-like". The Advantage of radio buttons is that all the options are visible right away, but the disadvantage is that they take considerable space at the GUI.

2. Use a drop down list with or without a "title". It takes less space than radio buttons. Slightly less fun to interact with but not too bad at all either IMO.

The advantage of those two approaches is that they don't try to fit our outcomes into a two bits "model", but rather relate to them as 4 independent options. Clearly three of them are in the same "domain", but we don't have combinations which behave weirdly anymore, and it actually gives information which otherwise users would have to look up at the manual - the xterm one is clear what it tries to be.

What do you think?

This PR? later? never?

mintty:
A/B/E/F, yes.

> I had two ideas for alternative control scheme, ...

I had another idea; instead of "Allow blinking", an "xterm" checkbox, and some "magic" interworking between the checkboxes. However, for `-o FontMenu=0`, the blinking checkbox is gone already and no space left...
Anyway, let's postpone that.

avih:
> However, for -o FontMenu=0

Huh, I did notice at the code there's an alternative font menu/dialog/something, but never tried that...

> Anyway, let's postpone that.

Yeah.

As for the docs, sounds good to me. This time I will leave the docs to you ;)

I'll update the PR today or tomorrow.

avih:
I think I'm good with it with these two commits. Now go crazy and add BoldAsRainbowSparkles :)

mintty:
> I think I'm good with these two commits.

Which means you've adapted the PR? I can probably check it only after the weekend though.

> Now go crazy and add BoldAsRainbowSparkles :)

Be careful with such remarks, I might actually do it :/ See the feature "separate colour for combining characters".

avih:
> Which means you've adapted the PR?

I have. Both are quite small patches in terms of changing code, but considering the long discussions we had, I thought it requires careful explanations, so that's where most of the time went...

> I might actually do it :/

How about, after some time off from bold, help me create good tools to measure performance of various aspects, and then we'll make it better where it matters? :)

avih:
Even if it's hopefully less surprising than before, it's still a complex subject, and I thought we owe the users a good explanation on how bold works.

if you only knew how much time I spent trying to understand the manual and document the behavior myself before I started looking at the code...

The funny thing is that the manual was totally correct in what it said, but it so conveniently failed to mention what happens when these options are disabled that it was maddening. You read the manual and it makes sense, you try it out and it's completely unexpected. You go back to the manual to see where it's different than the actual behavior  - and it's not wrong in anything it says.

Until I finally realized it's wrong in letting users assume one thing (that when the options are disabled it does the opposite of what it says they do when enabled), and then going behind their backs and doing a completely different thing.

We owe it to users. Hopefully this docs update will make it clear what happens.

mintty:
Your extra text seems OK (while the individual descriptions I proposed for the two options before seem clearer to me), but as you said yourself before, many people do not read the manual in the first place, so this is not the solution to all confusion...

avih:
> while the individual descriptions I proposed for the two options before seem clearer to me

It's almost verbatim copy of what it was before, just without the exceptions, and the "exceptions" are now explained hopefully clearly in a section of their own. Other than the exceptions, I thought the original text was quite OK.

> this is not the solution to all confusion...

Indeed. The solution is not to confuse users in the first place, where currently it's definitely non obvious when the two checkboxes are off, and IMHO before this PR it was even more confusing.

Maybe some day we'll change it to a dropdown list or radio buttons which won't have any exception.

Regardless though, the behavior is complex however you look at it - even if it was a dropdown list, so IMHO at the very least we should not shy away from a proper explanation, which is what I tried to do with the updated docs.

avih:
Thanks for merging it.

I see you added a line at the docs:
```
Note that Default bold display can be overridden by the BoldColour mode.
```

I've looked at the sources and I'm not quite sure what this mode is. Did you mean option rather than mode? If yes, it doesn't really override the bold-as-colour config, it just sets the colour which is getting used, the same as an OSC sequence can set it or an automatic colour is chosen if nothing set it.

mintty:
BoldColour mode is triggered by the BoldColour option if also enabled by BoldAsColour. It corresponds to xterm colorBD with colorBDMode; in this respect, BoldAsColour corresponds to two different xterm resources (the other is boldColors). The option value can be overridden by the OSC.

avih:
Is there a variable at the code or a config value which corresponds to this "mode"?

Otherwise, it's, IMHO, just an over complicated way to say:
- BoldAsColour sets a different colour to the bold text attribute, period.

The colour which BoldAsColour uses is the value of the BoldColour option if set - which could be overridden by an OSC escape sequence. If BoldColour is not configured and also an OSC sequence didn't set it, then an intense colour will be generated automatically.

But for the most part, if you set BoldColour, then that's the colour which will be used for the default colour with the bold text attribute.

Because the fact is, programs which use this OSC sequence are very rare.

mintty:
Also, BoldAsColour affects both ANSI8 and Default, while corresponding xterm modes are enabled by two different resource. I'm considering to change this (i.e. decouple BoldColour from BoldAsColour), so there should be no rewording now.

avih:
> BoldAsColour affects both ANSI8 and Default

Of course, because we think users will likely want to control these two together, and we don't want to confuse them with too many config values, then BoldAsColour controls different colour for the most common colours, and BoldAsFont controls emboldening of the most common colours.

Most common colours being ansi-8 and the default colour.

> while corresponding xterm modes are enabled by two different resource.

Of course, it should not be a combination of two config values, and we already suggested radio buttons or a drop down list/menu. Do you have other suggestions of a better control scheme other than radio buttons or a drop down list?

> i.e. decouple BoldColour from BoldAsColour

This I completely don't understand, and not for the first time. Can you please explain why you consider them "coupled"?

At the code the value of BoldAsColour is unaffected by BoldColour and vice verse. Really, they're completely unrelated values. Why do you keep referring to them as "coupled"? They're really not coupled at all.

Also, neither the code nor any comment nor the manual ever mentions anything related to "bold mode" other than an internal property of the font, i.e. how should emboldening of this font be performed from a technical point of view (real bold or shadow). xterm maybe has it, but mintty certainly does not have any documented or otherwise "bold mode".

The only thing which is non intuitive, and we know it, is the specific no/no combination which makes mintty behave like xterm. BoldColour is just fine and there's no need to do anything with it.

mintty:
> At the code the value of BoldAsColour is unaffected by BoldColour and vice verse.

No, if you start two sessions with BoldAsColour=no/yes and use my test file, you'll see BoldColour is applied only in the 'yes' case. If you agree it's OK to enable Default and ANSI8 bold colour modes with the same config, we can also leave it as is.

avih:
> No, if you start two sessions with BoldAsColour=no/yes and use my test file, you'll see BoldColour is applied only in the 'yes' case

Of course. It's exactly the same way that the BoldGreen value will only be used for `\e[32m` if you set BoldAsColour=yes. That's its goal at the config - you define your preferred colours once, and then can choose to use or not use them easily from the GUI.

BoldAsColour=yes means "use a different colour for the bold text attribute". If you don't request a different colour then of course the BoldColour and BoldGreen and all the other bold colours at the config are ignored. BoldColour only goal is to define a colour which gets used _if_ you configure it to use a different colour in the first place.

It's perfectly valid that some config values might not get used under some conditions.

This is perfect separation and it's by design. It doesn't matter what xterm does or doesn't do on such case. Mintty's behavior here is solid and just works.

Unless you can tell me why you consider this an issue, i.e. from a user perspective (most importantly, is there a reasonable thing the user might want to do but can't?), then I don't see an issue at all here.

mintty:
@avih, your review of the recent patch would be appreciated.

mintty:
There is this problem:
BoldAsColour is now combinable with BoldAsFont (which is wasn't before), and not only for ANSI8 but also fore Default. If there is no BoldColour defined, this means despite a bold font being used, mintty will also apply `brighten`. If brighten applies the "too close" branch to ensure contrast even in the opposite direction, the result is that the bold font appears to be dim. This problem existed already before, but obviously there are now additional cases that exhibit it.
grafik
grafik

I see two solutions:
1. If BoldAsFont is true, apply BoldAsColour only to ANSI8, not to Default.
2. Tune the brighten algorithm not to actually dim, at the cost of occasional low contrast.

As I'd like to release 2.8.2 next weekend, comments are appreciated.

avih:
I think this has nothing to do that AsFont can be used together with AsColour and everything to do with brighten being buggy.

The same issue of dimmed bold colour should have happened also before this PR when only BoldAsColour=yes and BoldColour is not set, so it uses brighten for the default bold colour, which sometimes ends up incorrectly dimmed.

So I don't see how the issue is a new one, but correct that now it can happen when AsColour=yes regardless of the value of AsFont, while previously it only happened when AsColour=yes and AsFont=no.

Am I wrong?

avih:
> I see two solutions:

No, please. Don't make it complex. That's the mother of all evil. It's simple. Brighten is broken, if you want to fix anything, fix brighten.

mintty:
I agree that changing impact of BoldAsColour would make things more complex,
but actually the issue here is that brighten is involved at all despite a bold font being used.
I think
      *pfgi |= 1;  // (FG|BG)_COLOUR_I -> BOLD_(FG|BG)_COLOUR_I
should be applied only if effectively no bold font is applied.

avih:
> but actually the issue here is that brighten is involved at all despite a bold font being used

Not at all. Nowhere did we say that if BoldAsFont=yes then colour should not be used. Recall, the whole goal of this PR was to make them independent. And now they are.

The issue is clearly that BoldAsColour doesn't work well for the default colour for some colours, which is obviously because brighten doesn't work well.

But it's not a new issue at all, and the default configuration also before this PR was equally affected by it.

I don't get why you refer to BoldAsFont at all when the issue is clearly with BoldAsColour and brighten.

avih:
Just leave it as is. There's no new issue, just another case where it can manifest. If you want to fix it, separate the different use cases of brighten such that each of them works perfectly for its use case, instead of using the same brighten function for different use cases where you have to make compromises to support several use cases.

mintty:
Hmm, not yet decided. BoldAsColour was never targeted to trigger the brightened faux-bold, see the old manual page.
I guess BoldAsColour should continue to enable BoldColour but otherwise be used only for ANSI8. Brightened bold should only apply as a fallback, if none of bold font, shadow, or ANSI8 is effective.

Independently of this consideration, I will have a look into brighten.

avih:
> BoldAsColour was never targeted to trigger the brightened faux-bold

It has. That's how we designed it. We said that the default and ANSI colours are the most common in use, therefore it would make sense to control them together. Controlling them making them using a different colour and/or emboldening them.

The exception is xterm mode where they apply bold differently.

Why do you try to find issues where there are not?

And seriously, the exact "issue" existed before the PR at the _default_ configuration. It's not a new issue issue at all, and everyone was able to bump into it by setting some default colour which brighten didn't work well with - just like now - no difference at all.

If you don't like it that BoldAsColour doesn't brighten correctly some values of default colour, then fix it. But don't make boldAsColour not affect the default colour because that was one of the main points of this PR. To allow both bold and colour - which was impossible before the PR. That's effectively the only visible thing which this PR adds.

Just leave it. Or make sure brighten works as expected on more cases than now. Don't add more logic of enabling and disabling stuff and making it behave unexpectedly just because you don't want to fix brighten.

mintty:
>>  BoldAsColour was never targeted to trigger the brightened faux-bold

> It has. That's how we designed it. 

That was a misunderstanding; I meant before the PR. But yes, we agreed on the current semantics, so I'll focus on the brighten effect as such. Thanks for the feedback.

avih:
> I meant before the PR

Before the PR it was documented like this:
```
By default, text with the ANSI bold attribute set is displayed with a
different colour, usually with increased brightness (mapping ANSI colours 
0..7 to their bright variants 8..15). This can be disabled here.

Note that when \fBBoldAsFont\fP is enabled, only bold text in one of the eight
ANSI colours has its colour changed, i.e. bold text without an explicitly
specified colour is shown with a bold font only.  This matches \fBxterm\fP behaviour.
Note also setting \fBBoldColour\fP.
```
Which means that by default - when BoldAsFont=no - it was also affecting the default colour. And if you go to the boldColour docs, it said this:
```
The colour to be used instead of the bold attribute can be configured.
The bold substitution colour is only applied if option BoldAsFont is false 
and option BoldAsColour is true.
The default is \fInone\fP, meaning that a brightened colour derived from 
the effective foreground colour is used.
The bold substitution colour can also be set and reset with the respective xterm OSC 
control sequences.
```

So yes, BoldAsColour=yes was designed to use the value of BoldColour or the faux brightened colour.

The only difference is that it was limited to when BoldAsFont=no, and now it's regardless of it.

mintty:
Here is option `BoldAsRainbowSparkles` for you :)

mintty:
Released 2.8.2.

PR #719, opened 2017-11-19
Copy to rtf: support true colours

avih:
> implement mapping of true colour → palette for RTF export

There you go.


mintty:
Great, thanks!

avih:
You're welcome.

PR #721, opened 2017-11-22
dim: simplify, improve, decouple from BoldAsColour

avih:
Quite simple, really...

### DIFF
--- a/docs/mintty.1
+++ b/docs/mintty.1
@@ -928,11 +928,6 @@ ANSI colours has its colour changed, i.e. bold text without an explicitly
 specified colour is shown with a bold font only.  This matches \fBxterm\fP behaviour.
 Note also setting \fBBoldColour\fP.
 
-This option also controls how the 'half-bright' (or 'dim') text
-attribute is displayed: if it is on, half-bright text is
-shown with halved foreground colour brightness; otherwise, it is shown
-by blending the foreground colour with the background colour.
-
 .TP
 \fBAllow blinking\fP (AllowBlinking=no)
 When text blinking is disabled, as it is by default, the blink attribute is
--- a/src/term.c
+++ b/src/term.c
@@ -1157,7 +1157,7 @@ term_paint(void)
 
           colour fg = win_get_colour(SEL_TEXT_COLOUR_I);
           if (fg == (colour)-1)
-            fg = truecolour(&tattr, tattr.truebg);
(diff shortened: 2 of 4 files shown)


avih:
Oh, actually, this also needs docs changes. Please wait with the merge.

avih:
> this also needs docs changes

Added docs, rebased, and force-pushed.

mintty:
I can do the docs, too, as you like.
However, this one actually withdraws some options. Especially, as I'd prefer to blend with current background, the removal of function truecolour may be premature.

avih:
> I'd prefer to blend with current background

This part of the commit message explains why it's a bug:
```
- Blend with the cell background only sort of works with non coloured
  background (gray/black/white). With coloured background the fg colour
  changes in an unexpected way for "dim".
```

And this bug currently (i.e. before the fix) already happens on all those cases (also at from the commit message):
```
If BoldAsColour=no and/or it's the default colour and/or it's a true colour:
- blend the fg with the cell's bg.
```

Let's have xterm as some reference:
Xterm default config: left side: top: normal, bottom: reverse (7). Right side: same as the left-side but also with dim (2).
xterm-all-combos

Now let's look at mintty with default configuration (so BoldAsColour=yes):
Left side: normal/reverse. Middle: with dim - with mintty before the PR. Right-side: with dim after this PR:
mintty-dim-fix-yes-colour

At the middle terminal (mintty before the dim change), and in it at the bottom output (for `sgrtab -g 7 2`). See how the first two lines look weirdly out of place? That's the default colour with and without bold - that's the bug.

Now compare it to the xterm dim at the right-side of the top screenshot.

The bug is also always affecting (not seen at the screenshots) true colours, And when BoldAsColour=no then it affects all colours.

For instance, the same three terminals with default everything except BoldAsColour=no:
mintty-dim-fix-no-colour

So now this weird behavior extended to everything.

It's a bug. It's exactly one of the bugs which this PR fixes.

avih:
> this one actually withdraws some options

For the record, if you look at the history of dim at the commit message, it becomes pretty clear it was never about a configuration option, it was a way to make dim "smarter" by taking more factors into account. And it only wanted to get smarter because it got it wrong in the first place.

It didn't end up very smart, certainly not after all the changes it went through, and the association between BoldAsColour and dim has always been a weird one to say the least, and eventually it "chooses" between one set of bugs to another set of bugs. Instead of just make it stupid and working and without being unnecessarily smart.

Issue #729, opened 2018-01-14
Option -B frame produces small title bar; unable to move window with Alt+Space.


R030t1:
When passing `-B frame` there is a small title bar on the top of mintty when on Windows 10. Is this intentional? I expected the window border and nothing more.

I am also unable to move the mintty window using the "Move" option on the dialog box produced by `Alt+Space`. The cursor teleports to the small area of title bar above the text area, but the cursor is the selection caret. Attempting to click and drag selects text. If I move the cursor, without clicking, to the main text area, the cursor changes to the movement cross. Clicking and dragging at this point produces no movement.

To avoid creating another bug: Is `-B void` supposed to include the scroll bar? Would that option be considered?

mintty:
This "title bar" is presumably part of the "frame" that you asked for. The options "frame" and "void" just configure some window style properties (see https://msdn.microsoft.com/en-us/library/windows/desktop/ms632600(v=vs.85).aspx), varying on WS_CAPTION, WS_BORDER, WS_THICKFRAME; I've also tried WS_DLGFRAME but that makes no difference. If this has the consequence that the window cannot be moved, that's out of control of mintty; ask Microsoft. However, the options were introduced to accomodate people who want a fixed and lean layout, so you would place the windows at your desired positions and not want to move them...

R030t1:
After some testing it seems like the behavior of mintty is seen in other programs, and is new with Windows 10. It looks like it will be hard or impossible to fix. I will comment again on whether or not I was able to work around the issue; the most likely candidate for a fix seems to be keeping the title bar, but drawing over it.

mintty:
Released 2.8.4.

PR #730, opened 2018-01-14
Add Ctrl+Alt+Click window movement.

R030t1:
In reference to #729; the window style chosen with `-B frame` allows no movement at all. The `Alt+Click` functionality works regardless of border style.

The code works by telling the desktop that the client area is actually the title bar.

If necessary I can work on adding the modifier to the options dialog. Where should it go? For Windows, any keys besides `Alt` may prove problematic due to unconfigurable UI behavior. `Shift` is also popular.

### DIFF
--- a/src/winmain.c
+++ b/src/winmain.c
@@ -2196,7 +2196,16 @@ static struct {
       break;
 #endif
     }
-  }
+
+    when WM_NCHITTEST: {
+      LRESULT result = DefWindowProcW(wnd, message, wp, lp);
+
+      if (result == HTCLIENT &&
+          (GetKeyState(VK_MENU) & 0x80) && (GetKeyState(VK_CONTROL) & 0x80))
+        return HTCAPTION;
+      else
+        return result;
+    }
 
  /*
   * Any messages we don't process completely above are passed through to


mintty:
Thanks for digging this out; 3 questions:
* Usually, DefWindowProc is only called as a fallback; does the handling need to be limited to the HTCLIENT case?
* I had checked GetAsyncKeyState vs. GetKeystate on another occasion before and concluded we'd better not use the Async variant; don't remember the reason, though. Can we do with GetKeystate at this point?
* Alt+click is already assigned to rectangular selection. Can we go with Alt+Ctrl+click?

mintty:
Please modify like this if you want it merged (or I could commit --author ...):
```
    when WM_NCHITTEST: {
      LRESULT result = DefWindowProcW(wnd, message, wp, lp);

      if (result == HTCLIENT &&
          (GetKeyState(VK_MENU) & 0x80) && (GetKeyState(VK_CONTROL) & 0x80))
        return HTCAPTION;
      else
        return result;
    }
```

R030t1:
Changed as requested.

* `DefWindowProc` is noted in the documentation to be called to handle any messages that you do not handle. There seems to be no requirement to call it last. Side effects are invoked through the return value, so I see no harm in moving it. In any case - there is no way to know the result of the hit test without running it, and the easiest way to do that is to use the existing implementation.
* I probably shouldn't have used `GetAsyncKeyState`. `GetKeyState` should be the default, but sometimes messages are not pumped quickly enough.
* `Ctrl+Alt+Click` is fine, another alternative is `Shift+Click`.

mintty:
Maybe I'm picky, but the first (empty) line has a trailing TAB, can you remove that please?

R030t1:
Fixed? I noticed you were using tabs to indent and was trying to keep that stuff sorted, my apologies.

R030t1:
Fixed? I noticed you were using spaces to indent and was trying to keep that stuff sorted, my apologies.

mintty:
There was a "}" line missing; fixed and added documentation.

mintty:
Released 2.8.4.

Issue #697, opened 2017-10-16
Aero snap not always working after moving between displays with different DPI scaling


Ede123:
Possibly related to #696:
After moving mintty between displays with different DPI scaling aero snap is broken in some cases (I often experience this issue as I grab mintty on my primary monitor, move it to my secondary monitor and then want to snap it to the side of this display).

Steps to reproduce with the setup described in #695:
* start with mintty on the primary (right) display with 100% scaling
* grab the window in the right part of the titlebar (e.g. close to the minimize button)
* move it over to the secondary (left) display with 150% scaling
* without releasing try to snap it to the left border of the secondary display
-> won't work (one needs to release and re-grab the window first

The same works the other way round:
* start with mintty on the secondary (left) display with 150% scaling
* grab the window in the left part of the titlebar (e.g. close to the icon)
* move it over to the primary (right) display with 100% scaling
* without releasing try to snap it to the right border of the primary display
-> won't work (one needs to release and re-grab the window first

Interestingly when grabbing the titlebar at the opposite end of what I described in these two scenarios Aero snap in fact *will* work.
An educated guess suggests #696 might cause this problem: As the cursor is moved off the mintty window, the mintty window has already left the active desktop area when the cursor hits the display border which probably causes Aero snap's implementation to fail.

mintty:
2.8.3 includes a tweak that is possibly related or could at least have impact (mintty/wsltty#80).
Does it fix the issue?

mintty:
Released 2.9.0 with support for DPI handling V2. Please recheck.

Ede123:
2.9.0 works nicely as far as I can tell. Thanks for fixing!

Issue #696, opened 2017-10-16
Window position moves relative to mouse cursor when moving between displays with different DPI scaling


Ede123:
Also visible in the screen capture of #695:
The cursor position changes relative to the window position when moving between displays with different DPI scaling. The "error" accumulates eventually moving the cursor completely off the mintty window when moving back and forth several times.

![mintty_2 7 9](https://user-images.githubusercontent.com/904570/31627455-a0d73c20-b2ad-11e7-8c7a-acf898610a65.gif)

mintty:
When rescaling to another DPI, mintty repositions itself. What the Windows mouse pointer does meanwhile, is out of control of mintty. Probably unfixable.

Ede123:
If mintty repositions itself in a way that maintains the windows relative position to the mouse cursor the issue would probably be solved.

I agree this might require some effort that might not be worth it, but I fail to see how this could be unfixable?

mintty:
Maybe my previous comment was a bit speculative as I don't have a dual-monitor Windows 10 system at hand. But whenever mintty rescales, it also calls a position fix win_fix_position

mintty:
Maybe my previous comment was a bit speculative as I don't have a dual-monitor Windows 10 system at hand. But whenever mintty rescales, it also calls a position fix (win_fix_position in winmain.c) in order to prevent it from reaching out of the current monitor. Maybe it would help to drop that, at a cost...

mintty:
Released 2.9.0 with support for DPI handling V2. Please recheck.

Ede123:
2.9.0 works mostly. There's seems to be one rare case where the mouse pointer still changes position relative to the window when moving between monitors, but I wasn't able to find any system behind it yet (or steps to reproduce reliably for that matter). When moving multiple times back and forth it happens sporadically.

Thanks for fixing!

mintty:
Released 2.9.5 with changed default HandleDPI=2.

PR #774, opened 2018-06-05
Add basic support for per-window DPI awareness v2, remove window flutter

v1ne:
This change adds basic support for per-window DPI awareness v2, introduced with Windows 10 1703.
In the long run, this enables mintty to use WM_GETDPISCALEDSIZE to resize the window pixel-perfectly when the pixel density changes so that the terminal's size in rows×cols stays constant.

For the moment, the basic support gets rid of mintty's window flipping back and forth between different screens A↔B if I drag it from a screen A onto a screen B with a different pixel density.

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -165,7 +165,7 @@ typedef struct {
   int geom_sync;
   int col_spacing, row_spacing;
   int padding;
-  bool handle_dpichanged;
+  int handle_dpichanged;
   int check_version_update;
   string word_chars;
   string word_chars_excl;
--- a/src/winmain.c
+++ b/src/winmain.c
@@ -198,11 +198,19 @@ load_library_func(string lib, string func)
 
 #define dont_debug_dpi
 
-bool per_monitor_dpi_aware = false;
+#define DPI_UNAWARE 0
+#define DPI_AWAREV1 1
+#define DPI_AWAREV2 2
+int per_monitor_dpi_aware = DPI_UNAWARE;  // dpi_awareness
 uint dpi = 96;
+// DPI handling V2
+static bool is_in_dpi_change = false;
 
 #ifndef WM_DPICHANGED
 #define WM_DPICHANGED 0x02E0
+#endif
+#ifndef WM_GETDPISCALEDSIZE
+#define WM_GETDPISCALEDSIZE 0x02E4
 #endif
 const int Process_System_DPI_Aware = 1;
 const int Process_Per_Monitor_DPI_Aware = 2;
@@ -214,6 +222,7 @@ DECLARE_HANDLE(DPI_AWARENESS_CONTEXT);
 #define DPI_AWARENESS_CONTEXT_UNAWARE           ((DPI_AWARENESS_CONTEXT)-1)
 #define DPI_AWARENESS_CONTEXT_SYSTEM_AWARE      ((DPI_AWARENESS_CONTEXT)-2)
 #define DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE ((DPI_AWARENESS_CONTEXT)-3)
+#define DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2 ((DPI_AWARENESS_CONTEXT)-4)
 static DPI_AWARENESS_CONTEXT (WINAPI * pSetThreadDpiAwarenessContext)(DPI_AWARENESS_CONTEXT dpic) = 0;
 static HRESULT (WINAPI * pEnableNonClientDpiScaling)(HWND win) = 0;
 static BOOL (WINAPI * pAdjustWindowRectExForDpi)(LPRECT lpRect, DWORD dwStyle, BOOL bMenu, DWORD dwExStyle, UINT dpi) = 0;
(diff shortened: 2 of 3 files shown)


mintty:
Thanks for this contribution which will need careful inspection.

Will this mitigate any of #695, #696, #697, maybe about the flipping effect?

About terminal size, please outline any benefit, also "in the long run", as compared to the current handling.

mintty:
Also (and this is actually one of the reasons of the project's "no unsolicited pull requests" policy), if this were a normal issue, its discussion would be more visible among the public issues. Should this be getting more complex, let's consider to move discussion to another issue.

mintty:
Ah, the context menu and the Options dialog scale properly, finally!

Biswa96:
Does this affect any old Windows versions?

mintty:
@Biswa96: I've tested it on Windows 7 and Windows XP, no side effects.

@v1ne: The names DPI_PROCESS_DPI_AWARE and DPI_PER_WINDOW_AWARE_V2 seem misleading to me. The previous mode (per_monitor_dpi_aware == true) was a per-monitor awareness, not per-process; also, checking https://msdn.microsoft.com/en-us/library/windows/desktop/mt791579%28v=vs.85%29.aspx, this and the new mode are both called _PER_MONITOR_, not per window.
Also, can we use #defines instead of enum Dpi_Awareness and stay with the name per_monitor_dpi_aware for now, so there are fewer or no changes in other functions and files in the first version? I've manually patched as attached, leaving in some old code (per #ifdef); after further evaluation and clean-up, I'd provide that again and give you the opportunity to submit a rebased patch if you desire (for the credits).


[winmain.c.txt](https://github.com/mintty/mintty/files/2080280/winmain.c.txt)

Biswa96:
Commits can be added to this pull request untill this is open.

mintty:
Not yet tested, but I assume the shortcut of win_fix_position is not proper; this function serves other purposes than DPI adjustment.

mintty:
Did you consider that some of the code you changed is disabled anyway?
    when WM_DPICHANGED:
#ifdef handle_dpi_on_dpichanged
...
#endif

mintty:
New integration proposal as attached, plus one "bool" -> "int" in winpriv.h
Evaluation of win_fix_position pending.
[winmain.c.txt](https://github.com/mintty/mintty/files/2081650/winmain.c.txt)

mintty:
I misinterpreted the win_fix_position patch, it only returns during the WM_DPICHANGED processing.
Still need to find a multi-monitor Windows 10 system for final testing.

mintty:
This is my cleaned-up proposal to integrate DPI handling V2.
I've tweaked the option HandleDPI so that you can enforce V2 (or nothing) but still use values 1 or true for backward compatibility.
Please resubmit as a (squashed) single-commit pull request. Alternatively, I may commit it next week.
[dpiv2.zip](https://github.com/mintty/mintty/files/2087693/dpiv2.zip)

mintty:
@v1ne, last reminder to resubmit the consolidated patch I uploaded, or reveal what I should use for `commit --author="..."`. Thanks a lot already for this contribution.

v1ne:
Hi. Sorry for the late reply. And sorry for the unsolicited PR. I'll stick to the policy in the future, sure!

I'm certain that this change fixes #695 and #696, which I experienced on a daily basis in my setup. I'm uncertain about #697.

Thank you for your helpful comments! I took your zip file and put it into the single commit. Indeed, this way I don't need to change unrelated files.

The benefit of using WM_GETDPISCALEDSIZE is that it allows mintty to maintain its size when the pixel density changes: Currently, if I move a 78×24 character terminal from my HiDPI to my LoDPI screen, it ends up at 81×26 characters. Without this message, it is difficult to achieve this without reintroducing the window flutter. At least from my experience of introducing HiDPI support in a major commercial application on Windows. But there, I have the advantage of not having to support HiDPI on Windows <10. ;)

If there's any other change you want, please tell me.

mintty:
You had dropped the man page from the commit, I've added that.
I would not have included the notes about possible future handling of WM_GETDPISCALEDSIZE in the commit message, but alright; do you have any specific idea how to make use of that?
Thanks again. If I had a name, I would list you on the credits page...

v1ne:
Thanks for merging! I don't use my name here on Github, sorry. 0:-)

I tried the following scheme:
`WM_GETDPISCALEDSIZE` with old dimensions → transform to rows×cols with current font (`cell_width`, `cell_height`) → use `win_init_fonts` to set new font, calculate new dimensions from rows×cols → hand new dimensions to Windows.
Unless the size of a cell changes linearly with the pixel density, I'd have to use the mechanism behind`win_init_fonts` to determine the new `cell_width`, `cell_height`. But I found the side effects of `wind_init_fonts` hard to control and didn't get an easy win this way.

Also, the size of the window decoration, search bar and maybe scroll bar have to be taken into account when calculating the window dimensions. I'd have to further look into how mintty's scroll and search bar are positioned to give this a second try.

mintty:
Released 2.9.0.

mintty:
Actually, it's just this change that causes DPI change handling to fail if the DPI of a monitor is changed,
even with option HandleDPI=1. Haven't compared with actual movement between monitors (because I don't have two at my Windows 10 system).
You noted in your commit message that something would still need to be done. Can we get this sorted out?

v1ne:
Can you give me a bit more context?

What I referred to is that on DPI changes, the new window size is currently computed only approximately, resulting in the window losing or gaining rows and columns.

mintty:
What I saw is: change DPI of the current (or only) monitor, and the terminal size jumps e.g. from 33x80 to 26x63, then after resetting DPI back to 32x79.
This is more than an approximate size.
Not yet tested whether it's the same when moving the window between monitors.
However, the fix is simply to maintain the terminal size by setting it to the previous one in the new DPI context, no need to precalculate it during WM_GETDPISCALEDSIZE. You may checkout the commit diff.

mintty:
Released 2.9.4.

mintty:
See #890 about reintroducing the manifest section `<dpiAware>true</dpiAware>` to support DPI scaling in Windows 7. This has become possible after introducing DPI awareness mode V2.

Issue #666, opened 2017-06-20
Can't set a background image


Bilge:
😞

mintty:
Your are welcome to propose an implementation.

Bilge:
& u

mintty:
Actually, this is a duplicate of #18, so I'm closing it.

thorstenkampe:
I propose an implementation based on the [background image functionality](http://www.9bis.net/kitty/?page=covidimus&zone=en) in `KiTTY`. `KiTTY` is a well known PuTTY enhancement. Since to my knowledge `mintty` was heavily based on terminal code from PuTTY in the beginning, it may be possible to incorporate that `KiTTY` patch into `mintty`.

The problem with that particular implementation seems to be the inability to display Hindi, Korean, Chinese, and Japanese characters. Greek, Cyrillic, and Vietnamese seem to be fine. Arabic better than in `mintty`.

mintty:
Released 2.9.0.

mintty:
Released 2.9.1. Image prefix '%' scales window to image aspect ratio.

mintty:
Released 2.9.4 with option `Background==`.

Issue #18, opened 2015-06-04
Background image


GoogleCodeExporter:
```
Some terminal emulators have a background image feature. This is firmly in
the realms of visual frippery though and unlike transparency it needs more
than a couple of function calls.
```

Original issue reported on code.google.com by `andy.koppe` on 2 Jan 2009 at 7:09

mintty:
Released 2.9.0.

mintty:
Released 2.9.1. Image prefix '%' scales window to image aspect ratio.

mintty:
Released 2.9.4 with option `Background==`.

elliott-beach:
Could you please provide an example of what the format of an absolute path looks like in the configuration file?

mintty:
Is https://github.com/mintty/mintty/wiki/Tips#background-image sufficient?

PR #775, opened 2018-06-08
Implements background bitmap & bitmap opacity support (#18, #666.)

lalbornoz:
Followup to bed2065c4f4dce5f7b81bad1e5332b6067fc4129.

This implementation considerably simplifies the original code, removing the (optional) dependency on GDI+ in particular, whilst support for tiling and non-BMP image types is presently not retained. The original implementation is, in my estimation, not terribly useful as non-transparent background images tend to make the actual terminal contents rather unreadable. Hence, a similar approach to KiTTY[1] and the much older `covidimus' PuTTY patch[2] is taken here instead, without overly bloating the code.

[1] <http://www.9bis.net/kitty/?page=covidimus&zone=en>
[2] <http://web.archive.org/http://www.covidimus.net/projects/putty/putty.php>

### DIFF
--- a/docs/mintty.1
+++ b/docs/mintty.1
@@ -1974,15 +1974,19 @@ The highlighting colours of search matches can be configured.
 \(en \fBSearchCurrentColour\fP=\fIbright yellow\fP
 
 .TP
-\fBBackground image or texture\fP (Background=)
-With this option, an image can be chosen as background.
-With an image file name, absolute (with an optional \fB~\fP prefix) 
-or relative to the current terminal foreground process, 
-the option either uses the image as a background picture, scaled to 
-the terminal size (optional prefix \fB_\fP to the filename), 
-or as a tiled background texture (prefix \fB*\fP to the filename).
+\fBBackground bitmap\fP (Background=)
+With this option, a bitmap can be chosen as background.
+With a bitmap file name, absolute (with an optional \fB~\fP prefix) 
+or relative to the current terminal foreground process,
+the option uses the bitmap as background bitmap.
 The background can be changed with an OSC 11 escape sequence, using 
-a filename prefixed with either \fB_\fP or \fB*\fP instead of a colour.
+a filename instead of a colour.
+
+.TP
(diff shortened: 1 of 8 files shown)


mintty:
Not having tested yet (thanks already anyway), a few comments about conditions and requirements:
* I think non-BMP support is essential from a user perspective, and this pulls in GDI+ because stubborn GDI doesn't load any non-BMP images. On the other hand, GDI+ brushes are apparently broken in Windows 10 (they do in Windows 7 and XP), so I was quite relieved to find out the final conversion method from GDI+ image to GDI BMP after having wasted, once again, much time in fighting the utterly insane Windows APIs.
* I also think tiled texture background is the main use case and thus essential. There are nice and moderate background textures which work well as background.
* A picture is in any case less useful unless it uses very moderate contrast. See the mintty homepage http://mintty.github.io/ for two screenshots. I'm not sure how transparency would impact the effect, please provide a screenshot.

mintty:
Looking at the code, you changed the background handling setup heavily, with no visible benefit. I see no reason for such a change, especially as essential functionality got lost. Also you seem to do some of the setup in the actual output code. There are a few lines of interest here, using a dedicated background DC, and the AlphaBlend function. If you provide a patch that is reduced to this improvement, and demonstrate its benefit with a screenshot, I'll consider that.

lalbornoz:
> [ ... ] with no visible benefit. I see no reason for such a change, especially as essential functionality got lost.
1) Code quality,
2) code quantity,
3) your code is useless unless you manually reduce any potential background image's contrast. Do you fancy making your terminal window unreadable much? I don't.
4) forget about it, I have better things to do than to have you `consider' anything.

mintty:
That's an unappreciated attitude in open source projects. I crafted a feature for some weeks, working heavily against Microsoft obtacles, resulting in somewhat chaotic but working code. You dropped essential features in a few days in exchange for a less useful partial feature and expect to get that accepted uncommented or what?
Thanks for closing.

PR #791, opened 2018-09-02
Honour zero value for SelectionShowSize

silverwind:
Before https://github.com/mintty/mintty/commit/4ff674fb586f6ad758cd74fe1c72399361556b7c, a value of `0` disabled this option which was no longer the case.


mintty:
Thanks for catching this. I wouldn't place it after initial calculations, though, but rather towards the top of the if case. Patch pending as I'm currently doing heavy changes in this file.

mintty:
(maybe just adding && cfg.selection_show_size to the if condition)

silverwind:
Updated with your suggested fix.

PR #799, opened 2018-09-19
loop-drain terminal input for less flickering and work

avih:
Previously we were trying to read once up to 4K input bytes, but
underlaying buffering could limit us. E.g. MSYS2 gives up to 256 bytes.

If an application outputs a bigger chunk, then it can result in partial
terminal/screen updates which typically manifest as flickering or
tearing or "not solid" screen updates etc.

This usually also meant more work because we will also update the screen
again once we read all the input.

Change it to drain the input in a loop as long as we have input, up to
the same 4K bytes as before.

This results is much less flickering of the status line in 'less' (and
therefore 'man' etc), and generally a more "solid" feeling on updates.

### DIFF
--- a/src/child.c
+++ b/src/child.c
@@ -455,13 +455,11 @@ child_proc(void)
 
     if (select(win_fd + 1, &fds, 0, 0, timeout_p) > 0) {
       if (pty_fd >= 0 && FD_ISSET(pty_fd, &fds)) {
-#if CYGWIN_VERSION_DLL_MAJOR >= 1005
-        static char buf[4096];
-        int len = read(pty_fd, buf, sizeof buf);
-#else
-        // Pty devices on old Cygwin version deliver only 4 bytes at a time,
+        // Pty devices on old Cygwin versions (pre 1005) deliver only 4 bytes
+        // at a time, and newer ones or MSYS2 deliver up to 256 at a time.
         // so call read() repeatedly until we have a worthwhile haul.
-        static char buf[512];
+        // this avoids most partial updates, results in less flickering/tearing.
+        static char buf[4096];
         uint len = 0;
         do {
           int ret = read(pty_fd, buf + len, sizeof buf - len);
@@ -470,7 +468,6 @@ child_proc(void)
           else
             break;
         } while (len < sizeof buf);
-#endif
         if (len > 0) {
           term_write(buf, len);
           if (log_fd >= 0 && logging)


avih:
Here's a sample video. Left side is "normal" mintty, right side is with the patch of this PR. Running `man`, then `less` (with some syntax highlighting), then tmux where one pane keeps writing output.

Sorry for the low quality video, but the status/cursor flicker are visible anyway. I tested that it plays in Firefox: https://0x0.st/sxRG.mp4

avih:
> More so, currently there's a bug ...

Added a first commit which fixes this issue. Now the buffer size can be set to any arbitrary size (e.g. 8K or 16K or 128K) without crashing while selecting with the mouse.

The patch to loop-drain the input remains identical as before (so still 4K).

mintty:
I was about to accept the first patch, but now about the second:
So if the length would wrap uint (which is very hypothetical), you would simply return, i.e. ignore the output?
And, as inbuf size might become quite large in this hypothecial case, maybe the reassignment should be guarded against failure of renewn?

avih:
> So if the length would wrap uint (which is very hypothetical), you would simply return, i.e. ignore the output?

Correct. As you noted, it's quite unlikely to happen, and all bets are off if it does, but at least we won't crash due to writing out of bounds.

Also note that it doesn't need the length to wrap. It "only" needs `term.inbuf_pos` to reach (a bit less than) `UINT_MAX / 4` - which is 1G. Still quite unlikely to happen but definitely possible to trigger in practice.

> And, as inbuf size might become quite large in this hypothecial case, maybe the reassignment should be guarded against failure of renewn?

I can protect against failure of `renewn`, but I don't think it's appropriate, for two reasons:
- It's not protected anywhere else, and protecting it only here would be largely meaningless.
- You might want to add some "central" protection, e.g. maybe at `renewn` itself such that it aborts mintty gracefully if an allocation fails. I.e. all callers can assume that `renewn` never fails.

But if you feel that protecting `renewn` here is appropriate, then sure, I can add it. Just let me know what you prefer.

mintty:
First of all, thanks for catching this potential overflow. I wasn't aware of this hold-back feature at all.
I've disabled it for testing and actually that isn't really harmful, so we could also drop it altogether and let output pass while selecting. It has some benefit though, but proper output should take precedence over uninterrupted selection. So if the buffer is about to overflow, we should rather just flush it out and not discard it. Changing to that policy would also make it appropriate to change to a fixed buffer (say, 4K also) rather than a growing one. If you grow it up to 1G you could expect other problems, too, anyway! (And in that case, also renewn shouldn't be a problem.)

avih:
> so we could also drop it altogether and let output pass while selecting

Hmm.. not very user friendly IMHO.

> So if the buffer is about to overflow, we should rather just flush it out and not discard it

This is fair IMO, though I don't see any issue with a growing buffer either.

Or you can combine both aspects and allow it to grow up to some reasonable size.. say 32K or some such, so it's a quadruple win - let the user select uninterrupted in most cases, doesn't use memory if there's no need, use as much as needed when needed, and also doesn't let it grow to unreasonable sizes.

Overall, I don't mind which option is taken, but I think it's be nicer to keep this feature in general (even if a fixed size).

Pick a solution. Once you do, I can write it if you prefer, or you do it. Just let me know.

Regardless, the patch `loop-drain terminal input for less flickering and work` is independent of suspend-on-selection as long as it doesn't increase the size beyond 4K.

avih:
Also, I think it should be similar to the case where the user scrolls up while input keeps coming (so the screen doesn't update but the scrollback buffer keeps growing), isn't it? how is it handled there?

mintty:
I tested scrolling to some extent, not too thoroughly yet. Feel free to test selection inside the scrollback buffer.

> the patch loop-drain terminal input for less flickering and work is independent 

Yes, but you put it together in one pull request and I don't see a github option to pull only one of the commits - is it possible?

mintty:
A dynamic suspend-output-on-selection buffer is already implemented, so let's keep that and set a moderate size limit. To suggest a limit, let me take my dice... 30800 bytes.

avih:
> set a moderate size limit

Hmm.. so I tried to do that, and it's easy to handle at the `term_write` side, but I got tangled up with releasing the selection state at `termmouse.c`. To be honest I'm not comfortable with making this change reliably myself.

If you prefer to do it yourself, go ahead, I'll wait. Else, I suggest to take my overflow fix patch which doesn't do anything other than avoiding a crash, and then at your leisure you can change it to whatever you want?

mintty:
> but I got tangled up with releasing the selection state at termmouse.c

I see no need to do anything there. The selection can stay active while output is processed. Just handle the buffer overflow and let it - well, overflow into the terminal, then we should be fine.

avih:
> well, overflow into the terminal

I don't like this approach, because it will result in earlier text (which got buffered while suspended during selection) to be shown an an incorrect later position.

I've removed the buffer overflow fix from the PR, and leaving it at this comment as reference. Feel free to take it or write something yourself based on it or however you like.

```diff
From 3f051b9e4893acede4fa2e508406d9da288b230b Mon Sep 17 00:00:00 2001
From: "Avi Halachmi (:avih)" <avihpit@yahoo.com>
Date: Tue, 18 Sep 2018 18:41:37 +0300
Subject: [PATCH] term_write: avoid potential buffer overflows

1. In the unlikely case where term.inbuf_size*4+4096 wraps-around uint:
   previously we wrote out of bounds, now we allocate enough to fit
   without wrap around, or ignore this input if we can't.

2. If term.inbuf_pos+len > term.inbuf_size*4+4096 : previously we didn't
   allocate enough and wrote out of bounds, now we allocate enough.
   E.g. first write is 5K (inbuf_size and inbuf_pos are still 0).

The second case currently doesn't happen in practice because term_write
is currently not called with len > 4096, but this could change, and now
the code works correctly for any arbitrary len.
---
 src/termout.c | 8 ++++++--
 1 file changed, 6 insertions(+), 2 deletions(-)

diff --git a/src/termout.c b/src/termout.c
index c4294f2..1c78809 100644
--- a/src/termout.c
+++ b/src/termout.c
@@ -2366,8 +2366,12 @@ term_write(const char *buf, uint len)
   * because the user will want the screen to hold still to be selected.
   */
   if (term_selecting()) {
-    if (term.inbuf_pos + len > term.inbuf_size) {
-      term.inbuf_size = max(term.inbuf_pos, term.inbuf_size * 4 + 4096);
+    uint min_size = term.inbuf_pos + len;
+    if (min_size < term.inbuf_pos)
+      return;
+    /* min_size verified to not wrap-around, but term.inbuf_size*2 might */
+    if (min_size > term.inbuf_size) {
+      term.inbuf_size = max(term.inbuf_size * 2, max(4096, min_size));
       term.inbuf = renewn(term.inbuf, term.inbuf_size);
     }
     memcpy(term.inbuf + term.inbuf_pos, buf, len);
--
2.19.0
```

mintty:
> >   well, overflow into the terminal

> I don't like this approach, because it will result in earlier text (which got buffered while suspended during selection) to be shown an an incorrect later position.

It shouldn't; to be tested. If that's an issue, a term_flush within the scrolling function might help.
I'll give that a try myself.

mintty:
I've uploaded a patch. Please check. It contains some debug trace output and the limit is set to 33 to enable testing feedback. I'll change that to 33333 and clean up later.
Test case in shell: `while true; do sleep 1; date; done`.
And we should rename the inbuf, which is very misleading, to output which it actually is.

avih:
> And we should rename the inbuf

Well.. it's an input buffer for `term_write` and is used in the context of `term_write`... (not that I care about the name, but whether a thing is an input or output depends on the context at which you look at it).

Also, not sure how these work:
```
#ifdef buffer_problems
# ifdef patch_799
```
Neither of them is defined or explained anywhere as far as I can see... I assume they're your private things so I'm going to remove these blocks for clarity (while looking at the code).

avih:
```c
    if (term.inbuf_pos + len > term.inbuf_size) {
      term.inbuf_size = term.inbuf_size + 10 * len;
      term.inbuf = renewn(term.inbuf, term.inbuf_size);
    }
```

This will write out of bounds (will not allocate anything) if `term.inbuf_pos + len` ends up bigger than UINT_MAX. At this function you don't have control of what `len` is, therefore it should deal gracefully with any `uint` value. And this code doesn't.

mintty:
I wasn't assuming len might ever be near UINT_MAX but you're right, we should secure this completely while at it. Thanks for the feedback.

avih:
Also, you could rewrite this:
```c
  if (term_selecting()) {
    if (term.inbuf_pos + len > term.inbuf_size) {
      term.inbuf_size = term.inbuf_size + 10 * len;
      term.inbuf = renewn(term.inbuf, term.inbuf_size);
    }
    memcpy(term.inbuf + term.inbuf_pos, buf, len);
    term.inbuf_pos += len;
    if (term.inbuf_pos < 33) {
      // hold back output in buffer and skip actual output for now
      return;
    }
    else {
      // flush output buffer
      term_flush();
      return;
    }
  }

  term_do_write(buf, len);
```

As this instead:
```c
  if (term_selecting()) {
    if (term.inbuf_pos + len > 33) {
      // flush existing output buffer
      term_flush();
      // then continue to write len normally below.
    } else {
      // hold back output in buffer and skip actual output for now
      if (term.inbuf_pos + len > term.inbuf_size) {
        term.inbuf_size = term.inbuf_size + 10 * len;
        term.inbuf = renewn(term.inbuf, term.inbuf_size);
      }
      memcpy(term.inbuf + term.inbuf_pos, buf, len);
      term.inbuf_pos += len;
      return;
    }
  }

  term_do_write(buf, len);
```

This doesn't solve the UINT_MAX issue, but at least doesn't grow the buffer if you're going to flush it now anyway.

mintty:
I lost overview how the original patch to child_proc was related to the later patches to term_write about the active-selection output suspending.
The latter causes a serious problem and as I don't seem to find a proper solution I tend to disable the output suspending at all because it's not really necessary.

avih:
well, I wasn't following which patches caused issues, but did you try my suggested patch above instead of the change you ultimately pushed instead?

mintty:
Your version does not make mintty stall.
But it flushes about every second line of output. I don't see benefit in suspending such small pieces.
Alternatives are:
1. Flush larger pieces; however, writing larger chunks to the client pty is probably the culprit of the stalling issue.
2. Discard large pieces; not really nice for a terminal.
I still feel inclined to disable this output suspension. It's not worth the trouble.

avih:
I meant the first version I posted, at this comment: https://github.com/mintty/mintty/pull/799#issuecomment-423682418

I'm assuming you tried my comments about your patch (where you only buffered 33 bytes), and yes, it's obvious this version would flush often. But my initial version shouldn't.

Regardless, personally don't feel strongly about this feature and don't really care if it's removed.

mintty:
The error is an embarassing wrong calculation of the size of memory allocation for the suspension buffer.
I'll fix that and make the buffer size configurable, including 0 which will disable the buffering.

mintty:
Released 2.9.4.

PR #828, opened 2018-11-30
Avoid nested functions

dscho:
It was reported to the oss-security mailing list in
https://seclists.org/oss-sec/2018/q4/82 that GCC 7 and 8 have known
issues generating code for nested functions, essentially voiding any
Data Execution Protection (DEP).

As it is unclear how quickly GCC can be fixed, let's just avoid the
nested functions altogether.

See https://github.com/git-for-windows/git/issues/1725 for the discussion on the Git for Windows issue tracker that triggered my working on this PR.

Note: this patch mainly moves the nested functions out of their previous
host function, if necessary adds a suffix to the function name, and in
several cases declares structures to be passed via the `lParam` argument
of the `Enum*()` functions, to retain access to the local variables of
the former host function.

Note, too, that trailing whitespace was preserved faithfully when moving
and de-indenting the code.

Also please note that this patch does *not* care about code that is not
compiled because it is hidden behind `#if ... #endif` guards.

Signed-off-by: Johannes Schindelin <johannes.schindelin@gmx.de>

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -2471,44 +2471,47 @@ font_size_handler(control *ctrl, int event)
   }
 }
 
-void
-list_fonts(bool report)
+struct data_fontenum {
+  HDC dc;
+  bool report, initial;
+};
+
+static int CALLBACK
+fontenum(const ENUMLOGFONTW *lpelf, const NEWTEXTMETRICW *lpntm, DWORD fontType, LPARAM lParam)
 {
-  HDC dc = GetDC(0);
+  const LOGFONTW * lfp = &lpelf->elfLogFont;
+  struct data_fontenum *data = (struct data_fontenum *)lParam;
+  (void)lpntm, (void)fontType;
 
-  int CALLBACK fontenum(const ENUMLOGFONTW *lpelf, const NEWTEXTMETRICW *lpntm, DWORD fontType, LPARAM lParam)
-  {
-    const LOGFONTW * lfp = &lpelf->elfLogFont;
-    (void)lpntm, (void)fontType;
+  if (data->initial) {
+    data->initial = 0;
+    if ((lfp->lfPitchAndFamily & 3) == FIXED_PITCH && !lfp->lfCharSet)
+      EnumFontFamiliesW(data->dc, lfp->lfFaceName, (FONTENUMPROCW)fontenum, (LPARAM)data);
+  }
+  else if (!lfp->lfItalic && !lfp->lfCharSet) {
+    if (lfp->lfFaceName[0] == '@')
+      // skip vertical font families
+      return 1;
 
-    if (lParam) {
-      if ((lfp->lfPitchAndFamily & 3) == FIXED_PITCH && !lfp->lfCharSet)
-        EnumFontFamiliesW(dc, lfp->lfFaceName, (FONTENUMPROCW)fontenum, 0);
-    }
-    else if (!lfp->lfItalic && !lfp->lfCharSet) {
-      if (lfp->lfFaceName[0] == '@')
-        // skip vertical font families
-        return 1;
-
-      wchar * tagsplit(wchar * fn, wstring style)
-      {
+    wchar * tagsplit(wchar * fn, wstring style)
+    {
 #if CYGWIN_VERSION_API_MINOR >= 74
-        wchar * tag = wcsstr(fn, style);
-        if (tag) {
-          int n = wcslen(style);
-          if (tag[n] <= ' ' && tag != fn && tag[-1] == ' ') {
-            tag[-1] = 0;
-            tag[n] = 0;
-            return tag;
-          }
+      wchar * tag = wcsstr(fn, style);
+      if (tag) {
+        int n = wcslen(style);
+        if (tag[n] <= ' ' && tag != fn && tag[-1] == ' ') {
+          tag[-1] = 0;
+          tag[n] = 0;
+          return tag;
         }
-#else
-        (void)fn; (void)style;
-#endif
-        return 0;
       }
+#else
+      (void)fn; (void)style;
+#endif
+      return 0;
+    }
 
-      /**
+    /**
 	Courier|
 	FreeMono|Medium
 	Inconsolata|Medium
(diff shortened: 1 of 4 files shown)


mintty:
There are lots of nested functions (and some very small ones) in win_key_down (wininput.c) which you didn't touch. (And, for the records, while I added some nested functions especially for enumerations and a few other ones, I did not put those in, they were already in use before I adopted mintty, maybe even since putty times (to be checked).)
Did you check it's not needed to move those out? Is their handling by gcc any different from other nested functions? Is there a simple way to check this (other than checking the compiled assembly code)?

dscho:
I identified the code paths that needed to be changed by compiling with `-Werror=trampolines`, as suggested on the oss-security list. With this patch, it compiles clean.

Maybe not *all* nested functions are handled via trampolines?

dscho:
Oh, for the record, this is the gcc version I used (MSYS2):

```
$ /usr/bin/gcc --version
gcc (GCC) 7.3.0
[...]
```

Biswa96:
One query. The issue is in GCC 7 and 8. Shouldn't GCC programmers fix their code instead of others?

dscho:
@Biswa96 this was discussed over at the Git for Windows ticket.

Yes, you are right, in an ideal world.

Practically, it will take time, especially to convince the GCC developers who lived with that design for over 15 years and even warned about the DEP ramifications of using trampolines. It will take even more time to get this into a GCC release. And then MSYS2 upgraded to said GCC release.

I do not want to force Git for Windows' users to wait that long. I really don't.

dscho:
@mintty anything more you need from my side?

Biswa96:
BTW, the nested functions becomes a feature of GCC. https://gcc.gnu.org/onlinedocs/gccint/Trampolines.html

mintty:
@Biswa96: compiling mintty with `make CCOPT="-fno-trampolines -Wtrampolines"` still raises the error.

mintty:
@dscho: Apparently, trampolines are only used if the nested functions accesses local data of the embedding function. The reason may be that in this case, the trampoline mechanism (assumedly similar to what is called closure elsewhere...) would achieve that the nested function can even still be used after the embedding function returned. This is, however, not needed in most use cases. On the other hand, for a C compiler, it may be hard or impossible to determine the difference.
Taking a look at win_synctabs (winmain.c line 828): If I copy the parameter initially:
```
static int _level;
_level = level;
```
and then a few lines further down, inside the nested function, refer to _level rather than level, it works without trampoline.
Maybe this leads a way to achieve our goal with more moderate modifications. I'll check your patch.

dscho:
@mintty did you mean to direct that `CCOPT` comment to me?

dscho:
> If I copy the parameter initially:
> 
> ```
> static int _level;
> _level = level;
> ```
> 
> and then a few lines further down, inside the nested function, refer to _level rather than level, it works without trampoline.

Of course, because then `_level` is not a local variable, but a `static` one.

I tried to avoid that because I am so not familiar with the source code and therefore have no idea where there might be multi-threading pitfalls lurking around the corner. The way I converted it is the way that the Win32 API is used elsewhere, building on the tradition of C compilers without nested functions ;-)

So while I believe you that you could come up with a more minimal patch that would appear to work, I would argue that it is a lot easier to reason about the correctness of the patch if it keeps the scope of the variables (mostly) intact: rather than replacing function-local variables by file-local ones, the current approach replaces them by fields in a struct that is passed from function to function to function.

I hope that you take this argument to heart. In the past you have been a bit Teflon about my arguments, and it did not really help our interaction.

mintty:
The CCOPT comment was addressed to @Biswa96, saying that I checked the -fno-trampolines option and it did not help.
Don't worry, although the other issue is closed now, I see the use case for this change and it's not about any dark areas of Windows APIs which had made me cautious about changes, not reluctant to arguments in a former case. Will need a few days of course. And actually, I guess we share a bit the traditional rule to never change a running system, considering also the wrapper, which could quite easily be fixed for the -w max issue...

dscho:
> Will need a few days of course.

I would be interested in getting in a patch on Git for Windows' side early, as we are really close to Git for Windows v2.20.0 (probably sometime early this coming week). So any indication whether you throw out this PR in favor of your changes (that would extend the scope of the variables, opening the code to multi-threading issues) or whether you are inclined to take it, would be helpful in determining what route to take on my side.

> I guess we share a bit the traditional rule to never change a running system

Right. If it is buggy, it is of course not really running, and needs change.

> considering also the wrapper, which could quite easily be fixed for the -w max issue...

Isn't it frustrating how some users raise a wish and then are totally unwilling to lift even their little finger to have their wish fulfilled?

In this instance, I am quite willing to say that this person gets what they want as soon as they show a bit of an eagerness to also do something about it. We're not kindergarten teachers over here, after all, and they are grown-up users, not toddlers.

BrianInglis:
Unless you intend to permanently remove function nesting, which you may think a good approach, you may want to branch the unnested functions, so that the nested functions may later be regression tested, when the compiler bugs are said to be fixed. 
Not that I am against adding generalized call by name to C, as long as they also add the full generality of e.g. lists of anonymous constant values assigned sequentially to the index variable in for loops, and the other resulting niceties such as compiler-supported coroutines, continuations, and closures. ;^>

mintty:
I am inclined to take the patch, and maybe tweak it a bit in a post-merge commit, and check that soon.
However, there is some other ongoing work on mintty before it will be ready for a release again (Options dialog...). Trying to get that done within next week.

mintty:
@BrianInglis, the plan is to remove only those nested functions that lead to "trampolines". While I don't really like to be forced to such flattened design, it's only about a few functions, and I don't like branching at all.

mintty:
Almost done with checking.
There was a mistake in config.c, where the flag should actually not indicate the “initial” call but the recursion level, so the font set with options -o FontMenu=0 (built-in font menu) or -Rf (font listing) would become wrong; fixed.
And I folded some functions back in (also as preferred by @BrianInglis...) after the analysis that it's not the nesting that causes trampolines but the usage of outer local data.

Reverted some functions that do not generate a trampoline:
winmain:sort_tabinfo/comp_tabinfo
winmain:update_tab_titles

Reverted some functions, keeping the change to pass on the parameter, 
so then they do not generate a trampoline anymore:
winmain:refresh_tab_titles
winmain:win_synctabs
winmain:get_monitor_info (parameter struct)
winmain:search_monitors (parameter struct)
wintext:findFraktur

Last (and biggest) change to be checked is adjust_font_weights, which I think I'll leave unfolded.

I can still do this as a merge-commit sequence if you prefer, or as a direct commit with a much smaller change footprint, as you like.

shoelzer:
@mintty Thanks for addressing this quickly. I just want to highlight one thing: [A comment on the Git for Windows issue](https://github.com/git-for-windows/git/issues/1725#issuecomment-443019983) noted that GCC can be configured to error when it would generate code that requires executable stack. Doing so would guarantee that all problematic functions are found and fixed in this commit, and also prevent future regressions. Enable this with `-Werror=trampoline`.

dscho:
> I can still do this as a merge-commit sequence if you prefer, or as a direct commit with a much smaller change footprint, as you like.

Why don't you force-push my branch?

mintty:
@shoelzer: done that already.
@dscho: because I've never done such a thing, being a minimalistic git user. I'll pull this and then update.

Biswa96:
@mintty Another query. You've added `ifneq ($(platform), msys)`. Shouldn't that code check "if the compiler is gcc" instead of "if the platform is msys"? May help: https://stackoverflow.com/a/28166605

mintty:
I could have checked the gcc version instead. This is simply to disable the option on an old system (which is still supported out of nostalgia) which does not yet have it.

mintty:
Released 2.9.5.

PR #841, opened 2019-01-10
Create .gitignore

Biswa96:
Add .gitignore file to mamage git folder easily. With this file `git add` will ignore the binary files. Hence developers can build mintty in cloned folder.


mintty:
Sorry, I'm not in favour of such a paradigm. There is no need to exclude anything from git add if you add only files that shall be added. Something like "git add *" is a bad idea anyway.

Biswa96:
OK, as you wish.

PR #844, opened 2019-01-29
Updated Spanish translation

jsanjose:
I've updated the Spanish translation for the latest version.


mintty:
Gracias.
Just a comment:
  msgid "&Shift"
- msgstr ""
+ msgstr "&Shift"
would not be needed, as identity is the default transformation.

PR #864, opened 2019-04-01
fix typo

srlehn:


mintty:
Thanks for caring, but that was actually on purpose, see https://www.dictionary.com/browse/role

PR #865, opened 2019-04-06
BellInterval directive: quash bells for this many msec after the last

naknak-:
Rapid consecutive bells in mintty can cause a harsh buzzing sound.

The effect is present whether using system sounds, a WAV file, or `BellFreq`/`BellLen`.  You can observe it (well, hear it) with this shell command that produces many bells:  `printf "\a%.0s" {1..100}`  

This short patch adds a config directive, `BellInterval`, which dictates the minimum time, in milliseconds, between bells.  Bells triggered during the post-bell silence period are ignored.

NB: bells are played asynchronously, so this interval includes the duration of the bell itself.  In other words, if the bell sound is itself longer than `BellInterval`, you'll get the old behavior with no guaranteed silence.

### DIFF
--- a/src/winmain.c
+++ b/src/winmain.c
@@ -1550,7 +1550,7 @@ win_bell(config * conf)
          unsigned long now = mtime();
 
   if ( (conf->bell_sound || conf->bell_type) &&
-      ((unsigned long)conf->bell_interval <= now-last_bell) ) {
+      ((unsigned long)conf->bell_interval <= now - last_bell) ) {
     wchar * bell_name = (wchar *)conf->bell_file;
     bool free_bell_name = false;
     last_bell = now;
--- a/src/winmain.c
+++ b/src/winmain.c
@@ -1546,9 +1546,14 @@ flash_border()
 void
 win_bell(config * conf)
 {
-  if (conf->bell_sound || conf->bell_type) {
+  static unsigned long last_bell=0;
+         unsigned long now=mtime();
+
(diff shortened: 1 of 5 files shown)


mintty:
I thought about such a feature when I worked on bell configuration but had not considered it worth the effort. But then, in fact with many bells mintty can be blocked for a while, also the effort isn't really big...

Patch format:
Please apply space around operators and assignment as elsewhere (4 times).
Please do not change the format of the sentence "A simple frequency beep..." on this occasion, which is not a distinct bullet item.
You may optionally drop the html file from the patch as it is auto-generated on release.

mintty:
It's funny that when you paste the test statement into the terminal with the mouse, you'll hear the bell(s) before the command is echoed. So I'll add a 'do_update();' at the beginning of win_bell.

src/winmain.c
```diff
@@ -1546,9 +1546,14 @@ flash_border()
 void
 win_bell(config * conf)
 {
-  if (conf->bell_sound || conf->bell_type) {
+  static unsigned long last_bell=0;
```
naknak-:
```suggestion
  static unsigned long last_bell = 0;
```

naknak-:
fixed spacing; I only saw three places where it was missing so you'll have to give me a hint if there's a fourth.

mintty:
now - last_bell

naknak-:
got it, thanks for your help

mintty:
Released 3.0.1.

mintty:
Released 3.5.1.

PR #888, opened 2019-06-20
Support IME control sequences

k-takata:
[Tera Term](https://ttssh2.osdn.jp/index.html.en) supports some IME control sequences: TTIMERS, TTIMESV and TTIMEST.
See: https://ttssh2.osdn.jp/manual/en/about/ctrlseq.html

This introduces the same functions to mintty.

### DIFF
--- a/src/termout.c
+++ b/src/termout.c
@@ -2075,6 +2075,7 @@ do_csi(uchar c)
   if (arg1 < 0)
     arg1 = 0;
   int arg0_def1 = arg0 ?: 1;  // first arg with default 1
+  static int ime_status = 0;
 
   // DECRQM quirk
   if (term.esc_mod == 0xFF && esc_mod0 == '?' && esc_mod1 == '$' && c == 'p')
@@ -2590,6 +2591,15 @@ do_csi(uchar c)
         insdel_column(curs->x, true, arg0_def1);
     when CPAIR('#', 't'):  /* application scrollbar */
       win_set_scrollview(arg0, arg1);
+    // IME control
+    when CPAIR('<', 'r'):
+      ImmSetOpenStatus(imc, ime_status);
+      win_set_ime_open(ime_status);
+    when CPAIR('<', 's'):
+      ime_status = ImmGetOpenStatus(imc);
+    when CPAIR('<', 't'):
+      ImmSetOpenStatus(imc, arg0);
+      win_set_ime_open(arg0);
   }
 }
 


mintty:
> See: https://ttssh2.osdn.jp/manual/en/about/ctrlseq.html

I don't see much description there. What's the use case?

k-takata:
Here is an example that controlling IME from Vim:
https://ttssh2.osdn.jp/manual/en/usage/tips/vim.html#IMEcontrol

This setting is very useful for CJK users.
This turns off the IME when leaving the insert mode and restore the IME status when entering into the insert mode.

mintty:
Confused: How would you change insert mode inmidst entering a multi-key character sequence? And would you really want to continue such an IME sequence after interrupting it and doing something else?

k-takata:
For Japanese users, IME should be on when input Japanese characters, and IME should be off when input ASCII characters. So we normally turn off the IME in normal mode, and turn on the IME in insert mode only when we want to input Japanese characters. To turn on and off the IME, we use a special key: normally "Half width/Full width" key.
So the following is a typical use case to edit Japanese text without this PR:

1. Start Vim.
2. Type `a` to enter the insert mode.
3. Type `Half width/Full width` key to turn on the IME.
4. Input Japanese text.
5. Type `Esc` to go back to normal mode.
6. Type `Half width/Full width` key to turn off the IME.
7. Execute some normal commands.
8. Type `a` to enter the insert mode.
9. Type `Half width/Full width` key to turn on the IME.
10. Input additional Japanese text.

With this PR and with the above setting, this can be simpler:

1. Start Vim.
2. Type `a` to enter the insert mode.
3. Type `Half width/Full width` key to turn on the IME.
4. Input Japanese text.
5. Type `Esc` to go back to normal mode. The IME state will be saved, and then it will be automatically turned off.
6. Execute some normal commands.
7. Type `a` to enter the insert mode. The IME state will be automatically restored.
8. Input additional Japanese text.

This reduces the needs for typing `Half width/Full width` key many times.

mnishz:
The most annoying thing I think is that some keystroke is stolen if I forget typing `Half width/Full width` key (0xf3 or 0xf4) after going back normal mode. (And I often forget to type...)

ime
It shows that `j` or `k` (move command) doesn't work when I forget to type `Half width/Full width` key.

I can see a lot of Japanese articles that introduce the way to disable IME automatically when leaving Vim's insert mode.

mintty:
Thanks, understood. I'll merge the patch and refactor it to move the Windows invocation into winmain.c and also add a description for wiki/CtrlSeqs.md.

mintty:
Please check my modified version.

k-takata:
Thank you. It works fine!

mintty:
Released 3.0.2.

PR #912, opened 2019-09-15
add dark mode support partially (mintty/wsltty#157)

Biswa96:


mintty:
Thanks a lot. Uploaded minor tweak and changelog.

BrianInglis:
How does this interoperate in Windows "Night Light" mode?

Biswa96:
Does that mode change anything to Command Prompt?What do you want to look mintty to be in night mode? AFAIK, the night mode is a UWP thingy.

BrianInglis:
On my monitor, everything including the background wallpaper and windows look like reddish-yellowy faded paper.

mintty:
What is night light mode (I remember something like it back in Windows XP...); how do I set it up for testing?
How does mintty currently look in this mode and why is it not satisfying?
And how many other specialties will there be that could be considered...?

BrianInglis:
Win10/Settings (Action Centre)/Night Light icon. 
Appearance is currently okay. 
Should also be tested with dark mode.

Biswa96:
The undocumented function `ShouldAppsUseDarkMode()` uses `HKCU\SOFTWARE\Microsoft\Windows\CurrentVersion\Themes\Personalize\SystemUsesLightTheme` registry. So, it **may be** replaced with `RegGetValueW()` to query that REG_DWORD value.

PR #919, opened 2019-09-27
Update Japanese translation for 3.0.3

k-takata:


k-takata:
Is it possible to make the items in "Options -> Text -> Emojis -> Placement" translatable?
It would be nice if the items in "Emojis -> Style" are also translatable.

mintty:
Placement values are now translatable.
Emoji styles are actually proper names (of the respective emoji sets), not descriptions.

mintty:
Released 3.0.4.

PR #957, opened 2020-01-12
zh_CN: Complete and optimize existing translations

GeekDuanLian:


mintty:
Thanks. Actually, you could have remedied the search arrow issue via localization

 msgid "◀"
-msgstr "◀"
+msgstr ""

GeekDuanLian:
> Thanks. Actually, you could have remedied the search arrow issue via localization
> 
> msgid "◀"
> -msgstr "◀"
> +msgstr ""

Yes, but I prefer the source to change, localization can only translate meaning but not change meaning, at least I think so.

PR #976, opened 2020-03-14
Create sv.po

Regnander:
For the help command I had to remove one space in the beginning of each line to make the text fit the within the window.


mintty:
Thank you. You could also make the window a bit wider (msgid "100", msgstr "105"...).

PR #984, opened 2020-04-10
Enhanced dark mode.

### LINKED ISSUE #983: Enhanced dark mode in Windows 10
Code here needs to be modified for Windows 10 1903+:
https://github.com/mintty/mintty/blob/ee2edc877658817c0f6a852e415e9c3dc07f254d/src/winmain.c#L4943

`ShouldAppsUseDarkMode` should be replaced by `ShouldSystemUseDarkMode` at ordinal 138, according to issue [here](https://github.com/ysc3839/win32-darkmode/issues/3). I've confirmed that.

Also, this constant should be alerted:
https://github.com/mintty/mintty/blob/ee2edc877658817c0f6a852e415e9c3dc07f254d/src/winmain.c#L4948

According to pull request [here](https://github.com/microsoft/terminal/pull/4817), (I had decompiled the dll provided by the DWM team, and found that) `DWMWA_USE_IMMERSIVE_DARK_MODE` maybe `20` on newer systems, and we should try to set both(`19` and `20`) to `TRUE` to ensure. As for the order, the DWM team choose to try `20` at first.

Also, for dark title bar and menus, we should call function at ordinal 135(`AllowDarkModeForApp(TRUE)` in 1809 or `SetPreferredAppMode(AllowDark)` in 1903) and `FlushMenuThemes()` at ordinal 136. The declaration of these functions could be found [here](https://github.com/ysc3839/win32-darkmode/blob/master/win32-darkmode/DarkMode.h).

Berrysoft:
Fixes #983 

* Use undocumented api `RtlGetNtVersionNumbers` to get build number.
* Determine if it is 1903+, and use different apis.

Before (current one shipped with msys2):
Before

After:
After

### DIFF
--- a/src/winmain.c
+++ b/src/winmain.c
@@ -193,18 +193,9 @@ static HRESULT (WINAPI * pDwmSetWindowAttribute)(HWND, DWORD, LPCVOID, DWORD) = 
 static HRESULT (WINAPI * pSetWindowCompositionAttribute)(HWND, void *) = 0;
 static BOOL (WINAPI * pSystemParametersInfo)(UINT, UINT, PVOID, UINT) = 0;
 
-typedef enum PREFERRED_APP_MODE /* undocumented, use capital letters to fit api style of winapi */
-{
-  PREFERRED_APP_MODE_DEFAULT,
-  PREFERRED_APP_MODE_ALLOW_DARK,
-  PREFERRED_APP_MODE_FORCE_DARK,
-  PREFERRED_APP_MODE_FORCE_LIGHT
-} PREFERRED_APP_MODE;
-
 static BOOLEAN (WINAPI * pShouldAppsUseDarkMode)(void) = 0; /* undocumented */
 static BOOLEAN (WINAPI * pAllowDarkModeForApp)(BOOLEAN) = 0; /* undocumented */
-static PREFERRED_APP_MODE (WINAPI * pSetPreferredAppMode)(PREFERRED_APP_MODE) = 0; /* undocumented */
-static void (WINAPI * pFlushMenuThemes)(void) = 0; /* undocumented */
+static DWORD (WINAPI * pSetPreferredAppMode)(DWORD) = 0; /* undocumented */
 static HRESULT (WINAPI * pSetWindowTheme)(HWND, const wchar_t *, const wchar_t *) = 0;
 
 #define HTHEME HANDLE
@@ -274,8 +265,6 @@ load_dwm_funcs(void)
       pSetPreferredAppMode = 
         (void *)GetProcAddress(uxtheme, MAKEINTRESOURCEA(135)); /* ordinal */
     }
-    pFlushMenuThemes = 
-      (void *)GetProcAddress(uxtheme, MAKEINTRESOURCEA(136)); /* ordinal */
     pSetWindowTheme = 
       (void *)GetProcAddress(uxtheme, "SetWindowTheme");
     pOpenThemeData =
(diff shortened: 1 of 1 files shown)


Berrysoft:
I confirmed again and found that we *should* use `ShouldAppsUseDarkMode`. The reason why it doesn't work on Windows 10 1903+ is because that the value of `DWMWA_USE_IMMERSIVE_DARK_MODE` changed. Corrected.

mintty:
Thanks. 3 comments:
1. I'd prefer to retrieve the build number with HIWORD(GetVersion()), easier call and one less library to inspect. I know GetVersion is marked deprecated but I doubt they'll drop it and mintty uses it anyway.
2. I saw on https://stackoverflow.com/questions/53501268/win10-dark-theme-how-to-use-in-winapi that SetPreferredAppMode should be called prior to any window creation. Should we move it a few lines up or does it work there alike?
3. As SetPreferredAppMode is an upward-compatible repleacement of AllowDarkModeForApp (same "ordinal"), could we simply got with the latter unconditionally and save the reproduction of undocumented typedef enum PREFERRED_APP_MODE?

mintty:
I've tried it (simplified as suggested), works. FlushMenuThemes does not seem to be needed.
[dark.patch.txt](https://github.com/mintty/mintty/files/4461025/dark.patch.txt)

Berrysoft:
1. `HIWORD(GetVersion())` returns 9200 after Windows 8.1, which is not the right build version.
2. It works fine there, but also OK to move a few lines up. I'll do this.
3. It's OK to remove `PREFERRED_APP_MODE` and replace it with `int`. But I think they are not fully compatible because `sizeof(BOOLEAN)==1` and `sizeof(PREFERRED_APP_MODE)==4`.

Berrysoft:
Yes, it seems that `FlushMenuThemes` isn't needed... I'll remove it then.

mintty:
GetVersion is sensitive to "manifestation". Mintty includes "manifest" entries up to Windows 10 (res.mft), so HIWORD(GetVersion()) reports the proper build number.
The build is not needed for this patch anyway if we continue with the SetPreferredAppMode signature only (which catches the previous case).

mintty:
To explain my hesitation about GetVersion: I don't like the idea to replace a function that is - even if marked deprecated - still available and documented by some undocumented function, somehow further supporting MS's policy of hidden APIs. If you don't follow here, I'll merge your pull request and tweak back on it, no big issue.

Berrysoft:
Well, I don't have a machine with Windows 10 1809, so I can't confirm that the signature of `SetPreferredAppMode` is valid on that. As the size of the parameters of the function may be not exactly the same, and `__stdcall` functions on 32-bit machine clears the calling stack itself, I'm not sure it will work properly.

I'll change `RtlGetNtVersionNumbers` to `GetVersion`.

Biswa96:
> Well, I don't have a machine with Windows 10 1809

You can create an "unofficial" ISO from here https://uup.rg-adguard.net/ and run it as a VM (unless you don't use WSL2).

mintty:
Tweaked / simplified, not using build dependency anymore.
Tested on older Windows build.
Thanks a lot.

mintty:
I've tried to extend the effect of dark mode to the Options dialog and search bar (Alt+F3), with unconvincing results. Could you have a look, please? Should something be done about `RegisterClass`?

Berrysoft:
I'm afraid that it coundn't be done without effort of Microsoft. I'll show what I've found, and it depends on you whether to darken the dialog and search bar:

* Button: set theme to "DarkMode_Explorer"
* Edit: set theme to "DarkMode_Explorer" and deal with "WM_CTLCOLOREDIT" message of the parent window.
* Dialog background: deal with "WM_ERASEBKGND" message

There's a [repo](https://github.com/ysc3839/win32-darkmode) attempting to darken some common controls, but it uses a lot of hacking and even so, it cannot darken the menu bar...

Since the dialog of File Explorer is in light theme, I don't suggest make dialog in dark mode.

mintty:
Thanks a lot. I've added some comments to the code, documenting what's darkened by the respective calls. And I've limited the darkening to the title bar of the Options dialog, maybe that should be done.

PR #986, opened 2020-04-30
Improve search performance.

account-login:
The search performance is low with a large scrollback. My "scrollback lines" option is configured to 50000 and mintty would freeze when I type in the search box. Also the scroll speed is reduced significantly when search highlighting is enabled.

The current implementation is frequently searching in the whole scrollback, which causes the performance problem. This PR will fix the performance problem by avoiding full-search.
A search region `[range_begin, range_end)` is maintained as the already searched region, the region is initially small, only covers the currently displayed region. As the terminal scrolls or the user presses the prev/next button, the search region expands on demand.

There is a small side effect introduced by this PR. Since we are not doing full-search, the "current" search result (highlighted in the orange background) may not be detected before the user presses the prev/next button or scrolls to the oldest line.

The PR also improves the case folding code, we can avoid the costly binary search by using a direct lookup table for ASCII characters, this reduces the search time a lot.

A large scrollback is very helpful because I often extract data from scrollback directly without re-running the previous command and redirecting output to a file.

The scroll speed of terminal matters because often the terminal is the bottleneck for the application running inside a terminal. You don't want your `tail -F` to be slow down by a terminal.
The scroll speed of terminal can be measured with a script: https://gist.github.com/account-login/3767fc3fbccae60905da14c9678cbd0d
The script will write lines to STDOUT while logging the output speed (lines per second) to STDERR.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -15,6 +15,7 @@
 # - TARGET: target triple for cross compiling
 #	values: i686-pc-cygwin, x86_64-pc-cygwin, i686-pc-msys, x86_64-pc-msys
 # - DEBUG: define to enable debug build
+# - DEBUG_OPT: define to enable debug build with optimization
 # - DMALLOC: define to enable the dmalloc heap debugging library
 #
 # The values of DEBUG and DMALLOC variables do not matter, it's just about
@@ -118,6 +119,8 @@ LDLIBS := -mwindows $(extralibs) -lcomctl32 -limm32 -lwinmm -lwinspool -lole32 -
 ifdef DEBUG
   CPPFLAGS += -Wno-trampolines
   CFLAGS += -g
+else ifdef DEBUG_OPT
+  CFLAGS += -fno-omit-frame-pointer -O2 -g
 else
   CPPFLAGS += -DNDEBUG
   CFLAGS += -fomit-frame-pointer -O2
--- a/src/search.c
+++ b/src/search.c
@@ -52,11 +52,36 @@ static struct {
 };
 #define case_foldn lengthof(case_folding)
 #define init_case_folding()
+
+static uint16_t case_folding_small[256] = {
+  0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21,
+  22, 23, 24, 25, 26, 27, 28, 29, 30, 31, 32, 33, 34, 35, 36, 37, 38, 39, 40, 41,
+  42, 43, 44, 45, 46, 47, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 58, 59, 60, 61,
+  62, 63, 64, 97, 98, 99, 100, 101, 102, 103, 104, 105, 106, 107, 108, 109, 110,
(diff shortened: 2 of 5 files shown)


Biswa96:
_Off-topic_, that type of bencmarking/testing scripts may be added in the repository to test if future commits speed thinga up or not.

Biswa96:
_Off-topic_, that type of bencmarking/testing scripts may be added in the repository to test if future commits speed things up or not.

mintty:
Thanks for the proposal (although unsolicited); comments:
* Output scrolling speed is not directly related to search/backward scrolling, this seems a bit confusing.
* The Makefile change is unexplained and does not seem related to the search change.
* We can speed up on ASCII case folding, but it would look simpler to just do it by range check and mask with 0x20; also 956 > 255 looks quite chaotic; also such a table should be arranged in groups of 8 or 16 entries.
* Folding out search stuff from term.c may make sense but should be a separate step then, to provide a clear view on the actual changes.
* You mention a problem with visible marks outside the previous search range (if I understood correctly); does your commit "retain search results when cycling search results" try to work around that?

BrianInglis:
ASCII is only 0-127 and speedups are only safe in that range - 128-255 is code page dependent and contains UTF-8 extension codes. 
IIRC mintty uses the current Unicode data to build tables as part of the (pre-?)build, so that approach should be followed in the PR, if mintty does not already do something to provide access to the data in CaseFolding.txt.

account-login:
Hi @mintty , thanks for taking the time for looking at this. Answers to your doubt:
- The "scrolling" mentioned the PR is caused by application printing something to the terminal, every time a new line is added, the whole scrollback is searched, thus the output scrolling speed is affected by whether there is a search pattern or not. You can verify this by running something that prints something endlessly, while the screen scrolls, open the search box and type something in the search box, the scroll speed should be reduced significantly if you have a large scrollback, and the CPU usage is high.
- The DEBUG_OPT variable added to the Makefile enables the build with both debug flags (assertion, symbols) and optimization flags on. This is helpful for testing.
- The range check for ASCII case folding is a good idea, do you want another PR for that, or append commits to this PR?
- As the "folding out search stuff from `term.c`", do you want another PR for that? I thought the search code is changed significantly and it does not help much for code reviewing.
- The commit "retain search results when cycling search results" has nothing to do with the problem mentioned in the PR. The problem mentioned in the PR is that `term.results.current` can not be determined (when the beginning of scrollback is not covered by the search region). The `term_clear_results()` inside `term_search_next()` does not affect the behavior of search, it only affects performance by avoiding unnecessary full-search. (A full-search happens when `idx` jumps to the beginning while the search region only covers the end and is far from the beginning, this can be avoided by throwing away the current search region.)

mintty:
> every time a new line is added, the whole scrollback is searched

Thanks for clarifying, I wasn't aware of that as I don't really use the search feature myself.
So I'd appreciate a new (or revised) PR that contains two commits, the folding out and the actual changes related to *this* optimisation. Leave out other optimisations and Makefile stuff for now please.

mintty:
Involving @kiwiz who contributed the search code and might have a comment.

mintty:
Released 3.1.5.

PR #988, opened 2020-05-02
improve search performance. (revised)

account-login:
https://github.com/mintty/mintty/pull/986 resubmitted.

### DIFF
--- a/src/term.c
+++ b/src/term.c
@@ -453,44 +453,39 @@ term_reconfig(void)
     term.vt220_keys = vt220(new_cfg.term);
 }
 
-static bool
-in_result(pos abspos, result run)
-{
-  return
-    (abspos.x + abspos.y * term.cols >= run.x + run.y * term.cols) &&
-    (abspos.x + abspos.y * term.cols <  run.x + run.y * term.cols + run.len);
-}
-
-static bool
-in_results_recurse(pos abspos, int lo, int hi)
-{
-  if (hi - lo == 0) {
-    return false;
-  }
-  int mid = (lo + hi) / 2;
-  result run = term.results.results[mid];
-  if (run.x + run.y * term.cols > abspos.x + abspos.y * term.cols) {
-    return in_results_recurse(abspos, lo, mid);
-  } else if (run.x + run.y * term.cols + run.len <= abspos.x + abspos.y * term.cols) {
-    return in_results_recurse(abspos, mid + 1, hi);
-  }
-  return true;
-}
-
 static int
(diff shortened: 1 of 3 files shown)


mintty:
Is this your search-3-pr branch? Asking for easier testing, as I need to fiddle this out against current work I'm doing on the same files, unfortunately.
Also I forgot to ask: does the __asm__ inline have significant advantage over C-based time retrieval? (See e.g. function mtime which is already available.)

BrianInglis:
Advantage of `rdtsc` over `mtime()` is low overhead returning approximate clock cycle counts +/- pipeline and speculation effects. 
For accuracy `rdtscp` should give more repeatable results with less uncertainty. 
If significant improvements are made, `mtime()` overhead is higher, but should be adequate to show differences: if time differences appear minimal, improvement is likely minimal. 

If rdtsc is required, it would be better to use `#include <intrin.h>` or `#include <ia32intrin.h>` with `__rdtsc()` `_rdtsc()` `_rdtscp()` macros, to invoke compiler and architecture builtins, avoid issues, and get compiler warnings if there are issues, across both 32 and 64 bit builds; see output from e.g.:
`$ fgrep rdtsc /{usr,lib/gcc/*86*/*}/include/**/*intrin.h`

**Also:**
It would be better if the initial patch avoided deleting complete functions, as it makes it less clear what is being changed and improved. 
A separate patch in the same series can clean up unused functions later (if required).

mintty:
Released 3.1.5.

kiwiz:
Thanks for pinging me. If I recall correctly, the search code is executed on every window update to handle the case where a window resize reflows text. Not sure if that is still a consideration from a quick read of the code.

mintty:
Thanks for the check.
@account-login, can you please check this case. If it is a problem, we might add an update when the terminal is actually getting resized.

account-login:
Thanks for the check. I've verified that the window resizing triggers search update as before, since the condition for search update is not changed.

kiwiz:
I went back to see why I implemented it this way and found this comment: https://github.com/mintty/mintty/issues/85#issuecomment-292057196

I don't fully understand what `result.idx` represents, but does the new approach obviate the requirement of refreshing results on every update?

mintty:
I do not fully understand either the previous or the current strategy, sorry; so checking some remarks of your referenced comment, I just checked Insert/Delete Line/Char which apparently works, so I have the impression we can go on with the current strategy...

kiwiz:
Sounds good, thanks all!

PR #989, opened 2020-05-02
simplify scrollback compression and improve performance.

account-login:
This PR optimizes the scrollback compression code, the aim is faster scrolling speed, lower CPU usage, and simpler code.

I've done some profiling and here is a CPU frame graph:
PerfViewData orig flameGraph1

A large portion of CPU time is spent on compressing the scrollback (`compressline`) when a new line is added.
The `compressline` uses RLE encoding for each field of `termchar`. Unfortunately, the RLE implementation is slow and has a large room for improvement.
The current RLE implementation is doing `memcmp` on encoded elements, which is slow because every element will be encoded once, this can be improved by doing comparison *before* the encoding, thus avoids unnecessary encoding.
The functions for encoding each element (`makeliteral_attr` etc) also causes slowness, because those functions are called in hot loops and are not inlined.

This PR introduces the following changes:
- For the encoding of `cattr` fields, RLE encoding is used. Besides the RLE optimization mentioned above, the encoding is in the form of "element + counter", which is simple to code than the original encoding.
- For `chr` fields, RLE encoding is overkill and not used, because the only compressible thing is the repetitive white spaces. non-space characters are encoded as-is, while white space character is followed by a counter.
- For `cc_next` fields, the encoding is similar to `chr`, only repetitive zeros are compressed.
- The whole array of `line->chars` is encoding in each pass, this saves the trouble of handling the linked list and simplifies code.

Here is the optimized framegraph:
PerfViewData 3 flameGraph2
The width of `compressline` (the second peak in graph) is reduced.

I've measured the memory usage of the optimized encoding, the result is on par with the previous encoding.

The compression code can be further simplified by replacing `makeliteral_attr`/`writevint` with a plain `memcpy`, at the expense of a small increase of memory usage. This is not included in this PR.

### DIFF
--- a/src/term.h
+++ b/src/term.h
@@ -205,6 +205,7 @@ enum {
 
 typedef unsigned long long cattrflags;
 
+// NOTE: cattr_eq() should be updated if new fields are added.
 typedef struct {
   cattrflags attr;
   colour truebg;
--- a/src/termline.c
+++ b/src/termline.c
@@ -49,6 +49,7 @@ struct buf {
   int len, size;
 };
 
+// Note: Inlining the add() function reduces the compression time by 30%.
 static void
 add(struct buf *b, uchar c)
 {
@@ -64,6 +65,34 @@ static int
 get(struct buf *b)
 {
   return b->data[b->len++];
+}
+
+/*
+ * Store the number, 7 bits at a time, least significant
+ * `digit' first, with the high bit set on all but the last.
+ */
+static void
+writevint(struct buf *b, int n)
+{
+  while (n >= 128) {
+    add(b, (uchar) ((n & 0x7F) | 0x80));
+    n >>= 7;
+  }
+  add(b, (uchar) (n));
+}
+
+static int
+readvint(struct buf *b)
+{
+  int val = 0;
+  int shift = 0;
+  int byte = 0;
+  do {
+    byte = get(b);
+    val |= (byte & 0x7F) << shift;
+    shift += 7;
+  } while (byte & 0x80);
+  return val;
 }
 
 /*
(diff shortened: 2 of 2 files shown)


mintty:
I don't see a significant problem with the compression code, except maybe that mintty could do without it. I'd prefer not to do substantial changes here, to avoid any risk.

mintty:
Closing this for low benefit/risk ratio. Thanks anyway.

BrianInglis:
When submitting PRs of any sort, especially this large, it is advisable to break them up into smaller, incremental patch commits. 
Each one should do a simple visible change that improves the outcome. 
The old and new code should be heavily commented to show what is being done, as well as a detailed commit log entry, with the first line a summary of the intent. 
If you are improving code and refactoring, it is easier to see what is happening if you either first improve the code in place, then refactor, or else refactor, then improve the code, one patch commit for each step, one factor at a time. 
I know it's a pain, but it makes it easy for the maintainer to see what you are changing and its effect, step by step, and increases the chances that your changes will be accepted. 
If you can use an editor like `gvimdiff` with three diff windows, you can view the original code, the working copy you are changing incrementally, and your desired target, pulling changes from the latter target into your working copy, saving, logging, committing, as you go: **BTDT**!
Sometimes it is easier to do this locally in git, creating a new `git commit -f COMMIT_MSG` for each patch, or the equivalent using e.g. `Tortoise-Git` to interactively view the incremental diffs and compose detailed log messages, then at the end `git format-patch --cover-letter` and `git send-email --compose --annotate` the patch series, so that the maintainer can review each and apply with `git am --interactive`, as is normal on mailing lists.

mintty:
Thanks, Brian, for this general advice.
However, please don't anyone feel encouraged to spend another lot of work on this right now, as I am not convinced that this would be a significant improvement at all.

BrianInglis:
Work, Corinna, and other project maintainers have trained me somewhat, but one gets interested, and before one knows it, one has hacked a few hundred lines. The effect is not obvious to others, whether it is an improvement or not. So you have to go back and take it slower in logical steps, and "show your workings", as they always said about experiments, problems, and proofs. 
This PR is currently too much of a mish-mash to see what it does, what it changes, how it fits, and its effects. 

If someone says there is an improvement, they have to put in the effort to make the case as obvious as the improvement, otherwise the improvement may not be worth making. 
If they can't half the time, or double the performance, no one is likely to notice any difference. One or more orders of magnitude is better! 
Benchmarks tend to demonstrate mainly that individual changes and differences are often fairly insignificant. 

The flame/frame graphics with truncated uninformative labels and no axes don't help: I'd prefer, and other may too, actual M/K/B frame sizes vertically, with time horizontally in m/u/nsec, as appropriate. 
Stripping mintty prefixes, showing the full function names, adding axes, ticks, units, and labels, perhaps on separate HTML or SVG pages, or anything doing better plots, might help. 

As would traditional top 10 or 50% of the functions counts and times before, after, and total, with both before and after (if different) top functions, showing before and after counts and times, so we might be able to see where most of the repetition is done and time is spent, and whether the amounts and changes are significant.

PR #991, opened 2020-05-06
Update german de.po, release Croatian hr.po

milotype:
Both translations are complete

PR #992, opened 2020-05-07
Croatian translation hr.po

milotype:
Translation with corrections:
- value for scale factor set to 130
- abbreviations for "&Sat:" and "&Lum:"
- changed some accelerators and some translations
This should be it … Thank you for your help.


mintty:
Thank you for this contribution. Closes #991.

PR #999, opened 2020-05-24
maximize child windows if the parent window is maximized.

account-login:
The window opened with `alt+f2` preserves window size, but not the window placement.
Pressing `alt+f2` on a maximized window opens a window with the size of a maximized window, but the opened window is not maximized, which looks strange.

### DIFF
--- a/src/child.c
+++ b/src/child.c
@@ -1143,6 +1143,14 @@ do_child_fork(int argc, char *argv[], int moni, bool launch)
     // propagate shortcut-inherited icon
     if (icon_is_from_shortcut)
       setenv("MINTTY_ICON", cs__wcstoutf(cfg.icon), true);
+    // provide environment to to maximize window
+    {
+      WINDOWPLACEMENT wndpl = {.length = sizeof wndpl};
+      (void) GetWindowPlacement(wnd, &wndpl);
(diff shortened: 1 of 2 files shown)


mintty:
Did you try in child.c whether it works with IsZoomed (which is used in mintty already)?

account-login:
Updated code to use IsZoomed.

mintty:
Thanks, this works nicely for maximized windows. It was actually on my to-do-list for a while already.
Now there's a difference between maximized and full-screen (Alt+Enter, no borders). With your patch, Alt+F2 on a full-screen terminal start a maximized clone, which is better than before, but should we clone it full-screen as well?
Holding your pull in case you'd like to tackle that too.
While going at it, two minor remarks in child.c: "to to" -> "to" in the comment, and single-statement conditionals typically go without braces in mintty - while that is not my preferred style, it's currently consistent style in the code. Maybe I'll change that one day...

mintty:
I've added the fullscreen mode, so I'm taking your patch and adding mine.

BrianInglis:
> single-statement conditionals typically go without braces in mintty - while that is not my preferred style, it's currently consistent style in the code. Maybe I'll change that one day...

Plans for gcc say that conditonal statements missing braces will be warned about in future releases.

mintty:
Released 3.1.7.

PR #1005, opened 2020-05-29
Support Windows 10 dark mode

adipose:


mintty:
The only difference I observe with this patch is that the background is dimmed (gray instead of black) when the window is unfocussed, independently of setting `Transparency=Glass`, while without the patch this behaviour is only caused by that option.

adipose:
And do you have w10 dark mode enabled?

mintty:
Yes, and title bars are dark. I do not really prefer that setting but I left it active for occasional testing...

adipose:
Something doesn't add up. Do you have accent colors enabled?

mintty:
No I hadn't, but it doesn't change much on quick testing.

adipose:
I also disable it.  What about 

HKCU\Software\Microsoft\Windows\CurrentVersion\Themes\Personalize\AppsUseLightTheme

adipose:
I fixed the bug I mentioned before, which was extending the client frame when glass was disabled.  Now that part is separated as well.

In the background:

image

In the foreground:

image

adipose:
I don't understand it, but now the original code is working fine.  Perhaps flipping windows10 theme back and forth within mintty fixed something.

PR #1043, opened 2020-10-15
tabbar: Don't recreate the tabbar every time

k-takata:
The height calculation in win_toggle_tabbar() was wrong. So, when the
function was called, the tabbar was recreated every time.

Because of this, changing the title sometimes took very long time.
(With the default installation of Cygwin, the current directory is shown
as the title. Thus, on my slow machine, executing a `cd` command
sometimes took more than one second.)

Fix the height calculation and remove an unused variable.

### DIFF
--- a/src/wintab.c
+++ b/src/wintab.c
@@ -10,7 +10,6 @@
 #include <string.h>
 
 int TABBAR_HEIGHT = 0;
-static int prev_height = 0;
 static HWND tab_wnd, bar_wnd;
 
 static HFONT tabbar_font;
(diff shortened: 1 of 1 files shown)


mintty:
While this patch cleans up height calculation a bit, I don't see how the previous calculation was wrong.
And particularly I do not see how this could change the creation of the tabbar, and affect speed.

k-takata:
Before this change:
```
height = cell_height + margin * 2;
TABBAR_HEIGHT = height + padding * 2 = cell_height + margin * 2 + padding * 2
```
So, `height` and `TABBAR_HEIGHT` can never be the same value.
Therefore, the following if-block is always executed when win_toggle_tabbar() is called.
https://github.com/mintty/mintty/blob/4022d5bac876bd64195099ac8dde84d5f3a9dfef/src/wintab.c#L289-L291
This means that the tabbar window is always recreated when win_toggle_tabbar() is called.
On a slow machine, recreating the tabbar takes long time.

mintty:
Released 3.4.1.

PR #1044, opened 2020-10-15
tabbar: Fix startup position

k-takata:
When tabbar is enabled, users may expect that all mintty windows are
placed at the same position.
If Alt+F2 is used, the new window is placed at the same position as the
current window. However, if mintty is executed in another way, the
position is not the same as the current window.

This fixes the position of a new window even if Alt+F2 is not used.


mintty:
Thanks for finding this tricky solution for this problem.
One issue remains: If I run 2 tabbar-enabled windows (say 1 and 2) from 1 non-tabbar window (window 0), window 2 will not align with window 1 but rather with window 0.
I have no idea how this could be handled too, and people who always run with tabbar will not care much, so let's ignore it.

mintty:
Released 3.4.1.

PR #1045, opened 2020-10-16
tabbar: Fix startup position of maximised window

k-takata:
This fixes two issues:

1. Distinguish a maximised window from a fullscreen window.
   This is an additional fix to #1044.
2. Fix Alt+Shift+F2 behavior on a maximised window.
   When Alt+Shift+F2 was pressed on a maximised window, the new window didn't become a maximised window.

### DIFF
--- a/src/child.c
+++ b/src/child.c
@@ -1015,7 +1015,7 @@ setup_sync()
       setenvi("MINTTY_DX", 0);
       setenvi("MINTTY_DY", 0);
     }
-    else {
+    else if (!IsZoomed(wnd)) {
       RECT r;
       GetWindowRect(wnd, &r);
       setenvi("MINTTY_X", r.left);
--- a/src/winmain.c
+++ b/src/winmain.c
@@ -5473,9 +5473,15 @@ main(int argc, char *argv[])
         (LPCTSTR)(uintptr_t)class_atom, NULL);
     if (wnd_other) {
       if (IsZoomed(wnd_other)) {
-        setenvi("MINTTY_DX", 0);
-        setenvi("MINTTY_DY", 0);
-      } else {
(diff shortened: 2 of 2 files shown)


k-takata:
BTW, I have a question about the behavior of Alt+F2.
Is the following expected?

1. Maximize a mintty window.
2. Press Alt+F2.
   A new window will open in a maximized mode.
3. End the maximized mode of the new window.
   The size of the window become almost same as the maximized size.

Shouldn't be the size of the window as the same as the default size (or the normal size of the previous window)?

mintty:
How is style WS_THICKFRAME involved in the logic?

k-takata:
If a window is in the fullscreen mode, the window doesn't have WS_THICKFRAME style.
If a window is not in the fullscreen mode (and if it is not executed with `-B void`), the window has WS_THICKFRAME style.

mintty:
I merged this as two separate commits, adapting the merge comment to reflect the two changes.
About WS_THICKFRAME, did you test the `-B void` case?

mintty:
About your other question (restore from cloned maximised): this is kind of "expected" by missing implementation of consistent behaviour. I guess, this could be achieved in child.c: setup_sync, indicating the maximised state (and maybe also fullscreen) separately, and setting the placement values to the restored ones instead of the current maximised ones.

k-takata:
> About WS_THICKFRAME, did you test the `-B void` case?

Yes. When `-B void` is specified, maximizing the window results in a fullscreen mode.
So, I think that this logic is correct.

k-takata:
> About your other question (restore from cloned maximised): this is kind of "expected" by missing implementation of consistent behaviour.

I will check if I can implement this.

mintty:
Released 3.4.1.

PR #1048, opened 2020-10-17
Fix window positioning

### LINKED ISSUE #1046: tabber: The height of tabbar is not considered when setting the window size
To reproduce:
1. `mintty -s 80x30 --tabbar`
2. The height of the terminal may become 28 rows. I expect it becomes 30 rows.

k-takata:
This fixes two issues about window positioning.

1. Fix Alt+F2 behaviour when the window is maximised.
   See: https://github.com/mintty/mintty/pull/1045#issuecomment-709704486

   This also changes the behaviour of `CSI 9 ; [0-3] t`.
   1. Maximise the window vertically.
      `echo -e '\e[9;2t'`
   2. Maximise the window.
      `echo -e '\e[9;1t'`
   3. Restore the window.
      `echo -e '\e[9;0t'`

   This didn't restore the original window size before maximised vertically.
   But now the original size is restored.
   (I'm not sure about the behaviour in other terminals, though.)

2. Consider the size of tabbar on startup.
   Close #1046.

### DIFF
--- a/src/child.c
+++ b/src/child.c
@@ -1140,8 +1140,8 @@ do_child_fork(int argc, char *argv[], int moni, bool launch, bool config_size)
 
     // provide environment to clone size
     if (!config_size) {
-      setenvi("MINTTY_ROWS", term.rows);
-      setenvi("MINTTY_COLS", term.cols);
+      setenvi("MINTTY_ROWS", term.rows0);
+      setenvi("MINTTY_COLS", term.cols0);
       // provide environment to maximise window
       if (win_is_fullscreen)
         setenvi("MINTTY_MAXIMIZE", 2);
--- a/src/termout.c
+++ b/src/termout.c
@@ -2493,7 +2493,7 @@ do_winop(void)
       // Ps = 9 ; 1  -> Maximize window (i.e., resize to screen size).
       // Ps = 9 ; 2  -> Maximize window vertically.
       // Ps = 9 ; 3  -> Maximize window horizontally.
-      int rows0 = term.rows, cols0 = term.cols;
+      int rows0 = term.rows0, cols0 = term.cols0;
       if (arg1 == 2) {
         // maximize window vertically
         win_set_geom(0, -1, 0, -1);
@@ -2506,10 +2506,11 @@ do_winop(void)
       }
       else if (arg1 == 1) {
         win_maximise(1);
+        term.rows0 = rows0; term.cols0 = cols0;
       }
       else if (arg1 == 0) {
         win_maximise(0);
-        win_set_chars(term.rows0, term.cols0);
+        win_set_chars(rows0, cols0);
       }
     }
     when 10:
--- a/src/winmain.c
+++ b/src/winmain.c
@@ -5479,8 +5479,8 @@ main(int argc, char *argv[])
   }
 
   if (cfg.tabbar && !getenv("MINTTY_DX") && !getenv("MINTTY_DY")) {
-    HWND wnd_other = FindWindowEx(NULL, wnd,
-        (LPCTSTR)(uintptr_t)class_atom, NULL);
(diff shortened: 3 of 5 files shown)


mintty:
Thanks a lot, the Alt+F2 / restore issue seems to be working with the patch. The startup size doesn't.

k-takata:
> The startup size doesn't.

Hmm, it looks to be working on my side.
Could you show the reproducible steps?

mintty:
Well, I seem to have done one of two test cases wrong.
Fresh invocation (e.g. from desktop shortcut) works.
Running `mintty -size ...` from a running terminal session adapts the size to the previous session, whether it's tabbar-enabled or not. But that's likely OK.
So it works, great.
Why is it marked "work in progress" and cannot be merged?

k-takata:
> Running `mintty -size ...` from a running terminal session adapts the size to the previous session, whether it's tabbar-enabled or not. But that's likely OK.

Oh, I forget to consider this case. I think that the size from the command line option should be adapted when tabbar is not enabled. Let me check this.

> Why is it marked "work in progress" and cannot be merged?

Because I was not confident about the change of `CSI 9 ; [0-3] t`.

k-takata:
> I think that the size from the command line option should be adapted when tabbar is not enabled.

It seems it works as expected.

mintty:
CSI 9 works fine for me.
Using `-size`, now you've talked about considering this, let's check this case:
TabBar not enabled in config.
Mintty running without tabbar, size 80x33. Start this:
`../bin/mintty -size 99x11` works fine.
`../bin/mintty -size 99x11 -tabbar` will capture the new terminal size and position over the old one (despite not having a tabbar), size is 88x33.
If you think this glitch is acceptable, we can go with it.

k-takata:
That might be a rare case, but I committed a fix for that.

Biswa96:
In `src/winmain.c`, there is FindWindowEx and FindWindowExA both. FindWindowEx maybe A or W depending on `-DUNICODE`.

k-takata:
That part doesn't use a string, so it's not affected by `-DUNICODE`.
Should I use A or W explicitly?

mintty:
This is an interesting point. I do try to maintain A vs W explicitly so mintty can be compiled with `make CCOPT=-DUNICODE` which I occasionally check. Calls without a string don't need the suffix.
However, in `(LPCTSTR)(uintptr_t)class_atom` there is a string cast involved.
As the atom thing was created with RegisterClassEx**W**, I think it may be safer(*) to use FindWindowExW here,
but then it is also necessary to adapt the cast to either (LP**W**STR) or (LPC**W**STR), dropping the 'T' which is a weird variant anyway; otherwise, mintty will not compile on 32-bit cygwin anymore.

(*) Why could that be safer? I had some very weird and very hard to debug interference when trying to enable Unicode support in the menu tree (left part) of the Options dialog. It was essential to changed something explicitly to 'W' in order to make that work, as commented in windialog.c, without any direct API relation, so that it could not be detected by compilation...

k-takata:
> (*) Why could that be safer? I had some very weird and very hard to debug interference when trying to enable Unicode support in the menu tree (left part) of the Options dialog.

In my understanding, RegisterClass and the message loop should use the W APIs to support Unicode text.
No need to use FindWindowExW if the window title or the class name doesn't include Unicode text.

mintty:
> No need to use FindWindowExW if the window title or the class name doesn't include Unicode text.

Both the title (albeit not appearing in RegisterClassExW) and the class (which does appear there) can have Unicode characters (see option `-o Class` for the latter, which can be used to group multiple tab sets).
I have tested both features work nonetheless, however, I'd just feel slightly more comfortable with the "W" version...

k-takata:
Updated

PR #1054, opened 2020-10-27
tabbar: Bring next window to top when closing

k-takata:
Without this change, the focus might go to an unrelated window when closing a mintty window.
This makes tabbar more like actual tabbed UI.

### DIFF
--- a/src/winmain.c
+++ b/src/winmain.c
@@ -3677,6 +3677,14 @@ exit_mintty(void)
 {
   report_pos();
 
+  // bring next window to top
+  if (sync_level()) {
+    HWND wnd_other = FindWindowExW(NULL, wnd,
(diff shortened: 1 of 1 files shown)


src/winmain.c
```diff
@@ -3677,6 +3677,14 @@ exit_mintty(void)
 {
   report_pos();
 
+  // bring next window to top
+  if (sync_level()) {
```
k-takata, line 3681: I'm not sure whether this condition is okay.
Should this be `>= 2` or `>= 3`? Or, should `cfg.tabbar` be also checked?

mintty:
Sync level 1 is described as "sync. position/size when switching/launching a session".
Closing is like launching, so I think it's OK. Thank you.

mintty:
Released 3.4.2.

PR #1149, opened 2022-03-26
Add Appveyor CI

Biswa96:


mintty:
Great, thank you!
Just a few questions before I take the patch:
Some repositories have .appveyor.yml files (with a leading dot), does it make a difference?
So "Visual Studio" ist just the name of their image that also contains cygwin, a bit confusing.
What does the initial line of the build_script section do (`- cmd: |`)?
The vim file has a separate test_script section, maybe that's a usual distinction.

Biswa96:
> Some repositories have .appveyor.yml files (with a leading dot), does it make a difference?

AFAIK, nope. Just a preference. Maybe `ls` would not show the file, that's why it is used.

> What does the initial line of the build_script section do (- cmd: |)?

It runs the commands in Command Prompt. [Docs]:

> AppVeyor allows you to choose between Command scripting and PowerShell scripting. If you are using the Command interpreter and batch files then prefix your script with `cmd:` as shown below.

[Docs]: https://www.appveyor.com/docs/build-configuration/

> The vim file has a separate test_script section, maybe that's a usual distinction.

OK, doing it right now.

mintty:
OK, much appreciated. This works.
I guess I may add a comment and also reduce the PATH to one extra component. Thank you.

Biswa96:
>  I may add a comment and also reduce the PATH to one extra component

I added it just for simplicity. Another way to do the build in a shell script. Then run the shell script in cygwin's bash. It may not require the PATH changes.

mintty:
I've tried to enable integration also for wsltty, which fails so far:
* the Appveyor build environment does not have Alpine WSL, and it does not have the `winget` tool so I don't see how it could be installed.
* the WSL it has does not have a build environment (no `make`)

Biswa96:
> the Appveyor build environment does not have Alpine WSL

But it has Ubuntu 😉 It may be possible to build the wslbridge-backend in a docker environment based on alpine.

Biswa96:
I have added appveyor CI build in wslbridge2. It uses alpine docker image to build the backend. If you can provide what the required steps should be I can try to add CI build in wsltty and probably in wsltty.appx also.

PR #1153, opened 2022-04-03
implement resetting of xterm and ANSI colours

ak2:


ak2:
I'd forgotten that the sequence for resetting the new separate ANSI values hadn't actually worked, and then noticed that resetting the xtern256 colours wasn't implemented, so I've done that, doing some refactoring in win_set_colours() along the way.

Further testing also showed that the code for parsing X colour names matched an empty string with the first entry ("snow"), so make sure the whole colour name is matched.

PR #1156, opened 2022-04-23
Improve legibility of blue and grey on light backgrounds

ak2:


ak2:
Made blue and grey (aka "BoldBlack") in the Helmholtz theme darker and dropped separate darker background colours for green and cyan, because they're more likely to be used as background for dark colours.

Also darkened the background variants of blue, red, magenta and grey a little more to improve readability of light text on them.

Changed the Kohlrausch colours accordingly, but kept brighter background colours variants there for readability of dark text on them.

Added copyrights.

PR #1162, opened 2022-06-16
Create monokai-dimmed

nshopik:
Add Monokai-dimmed theme

### DIFF
--- a/themes/monokai-dimmed
+++ b/themes/monokai-dimmed
@@ -1,0 +1,21 @@
+# source: https://github.com/gerane/VSCodeThemes/tree/master/gerane.Theme-dimmed-monokai
+
+BackgroundColour=31,31,31
+ForegroundColour=185,188,186
+CursorColour=185,188,186
+Black=58,61,67
+BoldBlack=136,137,135
+Red=190,63,72
+BoldRed=251,0,31
+Green=135,154,59
+BoldGreen=15,114,47
+Yellow=197,166,53
+BoldYellow=196,112,51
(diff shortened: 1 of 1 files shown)


mintty:
Thanks for the suggestion. The colour scheme looks sufficiently distinct from others to enrich the selection of bundled themes.
Two issues though:
There are a number of Monokais, why would you suggest the "dimmed" variant more than the others? Some motivation would be appreciated.
The reference you put in the theme file leads to a repository that looks like a private collection, raising potential copyright issues (note that there is a "Monokai Pro" plugin offered for purchase). I'd prefer the reference to point to https://iterm2colorschemes.com/.

nshopik:
Well its matter of prerenal preference i guess, but one thing i dislike about default monokai and others you link is that all have higher contrast (especially default white) and i found many other schemes in mintty aren't fit to my preference of contrast, thus why dimmed. I fond of dark solarized which i used long time for mintty as well.

BrianInglis:
It may be aging eyesight `;^>` or it may be sophistication `;^p` but I have moved my palettes to lower contrast colours and styles as they seem to render more distinctly against both bright and dark backgrounds, and lower contrast backgrounds better support rendering of higher contrast colours, where high contrast colours on high contrast backgrounds are either visible or invisible e.g. blue on black or yellow on white! 
My visual preferences are `DejaVu Sans Mono` 9 in `SkyBlue4` on `Beige` (or closest match) across all devices. (Also family tartan background wallpaper and lion rampant flag sign on screen where supported).

PR #1172, opened 2022-09-15
Update sv.po

Regnander:
The string "Opaque when focused" was a bit too long, which I solved by using a synonym.

| Before | After |
|-|-|
| Before | After |


mintty:
Thanks. The additional explicit changes
```
msgid "◄"
- msgstr ""
+ msgstr "◄"
```
are not needed as an empty msgstr will cause the msgid to be used.

Regnander:
I have removed them now.

PR #1176, opened 2022-10-17
Fix #1175: Line cursor invisible (zero sized) if cell_width is small enough

### LINKED ISSUE #1175: Line cursor invisible (zero sized) if cell_width is small enough
After fixes for #1157 the line cursor had became invisible (zero sized) if the cell_width is small enough (< 8).
The "offending" line is this one:
https://github.com/mintty/mintty/blob/7c90db2788107a3f68c044eacff01b6e1dce0f17/src/wintext.c#L4120
where, if cell_width is less than 8 then caret_width is multiplied by 0.
Maybe it would be better to ensure a minimum value (eg. 1) in case caret_width is zero at that point.

paolo-sz:


mintty:
Thanks, that would be
  caret_width = caret_width ?: 1;
in short style, but how about
  caret_width *= cell_width / 8 + 1;
or
  caret_width *= (cell_width - 1) / 8 + 1;
?

paolo-sz:
Well all three options behave slightly differently even if they behave the same for cell_width values between 1 and 7.

caret_width = caret_width ?: 1;
simply takes care of the special case where the result of the previous operation is zero without touching the main reasoning behind the previous calculation.
This is the best option from my point of view.

caret_width *= cell_width / 8 + 1;
behaves a lot differently in case of cell_width values between 8 and 15 where it multiply by 2 instead of 1 (and multiply by 3 instead of 2 in range 16 - 23 and so on with lighter impact as the value of cell_width increase).
This is not an issue a priori, but the caret_width result is no more linear with cell_width, the smaller cell_width, the bigger, in percentage, caret_width.

caret_width *= (cell_width - 1) / 8 + 1;
Unless cell_width could assume a value equal to 0, in my opinion, this is the worst option. It beaves as the previous case just shifting the ranges.

mintty:
OK, I'll take your patch then and tweak it to get the following instead, to consider high-DPI screens:
          if (caret_width < (int)dpi / 72)
            caret_width = dpi / 72;

paolo-sz:
Thinking better to other options like you did, there is another one that may be the best in case cell_width could be zero...

<code>caret_width *= (cell_width + 7) / 8</code>

this ensure a zero result if cell_width is 0 (could this be possible?), 1 in range 1-8, 2 in range 9-16 and so on

PR #1193, opened 2022-12-26
Add single macro to log info and errors

Biswa96:


Biswa96:
This is just an idea. I think this type of macro may improve the readability of the code.

I came up with the idea after seeing the commit https://github.com/mintty/mintty/commit/bb82bd2644e342c0266b77e0a08a58bade1775cf. Every source file has their own macro and commented printf statements. If there is only single macro present it may help to debug the issue quickly just by compiling with single `-DDEBUG` flag. Also this seems to improve the code readability as there is no various `#ifdef dont_debug...` macros and commented printf lines. The Makefile would get simpler. In case of any issue, the user can download or compile the mintty.exe with DEBUG enabled. And the output can be shared in issue details.

So, what do you thing about this change? If you don't like it's OK.

mintty:
Thanks for the suggestion. The debug__something_ pattern has the advantage of being more targeted than a single commone DEBUG flag. In some files (e.g. wintext.c) there are so many debug statements that I wouldn't like to need to enable all of them to debug a specific problem. E.g. I had to debug font issues a few times and was quite glad I could enable that selectively.
I is not necessary to change the source to debug; if you remove the binary .o (or touch the source), you could `make CCOPT=-Ddebug__whatever_` case by case. I often did that to enable tracing Windows event messages in winmain.c.
Speaking of tracing, sure code becomes quite cluttered with lots of #ifdef debug... lines, so for some specific ones I use trace_ macros instead, in turn enabled by a corresponding debug_. This is similar to your LOG but again, more selective.
To summarize, mintty's trace and debug approach is not perfect but does its job. Also, I think, for the third time I should try to continue mintty in a kind of maintenance mode now, so I'd prefer not to spend effort into just beautifying the code.
This is because, with the recent introduction of DEC status line, I believe mintty is meanwhile quite complete as far as terminal features are concerned, and hopefully not much more significant work needs to be added...

Biswa96:
I understand. Thank you for your review.

PR #1198, opened 2023-01-26
Update fr.po

Gozer-404:
Did a global review of the complete file today

### DIFF
--- a/lang/fr.po
+++ b/lang/fr.po
@@ -1,14 +1,14 @@
 # Traductions françaises du paquet mintty.
 # Copyright (C) 2017 THE mintty'S COPYRIGHT HOLDER
 # This file is distributed under the same license as the mintty package.
-# Eric lassauge <lassauge@gmail.com>, 2017-2020.
+# Eric lassauge <lassauge@gmail.com>, 2017-2023.
 #
 msgid ""
 msgstr ""
 "Project-Id-Version: mintty\n"
 "Report-Msgid-Bugs-To: https://github.com/mintty/mintty/issues/700\n"
 "POT-Creation-Date: 2022-12-18 01:37+0100\n"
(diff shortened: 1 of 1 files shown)


Gozer-404:
Sorry closed by mistake

mintty:
Thanks. Note that it is not necessary to add identical transforms (e.g. msgid "xterm" msgstr "xterm") as an empty msgstr will just default to the msgid.
Also, about the extended Sup->Super, Hyp->Hyper, please check whether they do not overflow their space in the Options menu, section Window.

Gozer-404:
Just a small comment on your note, I did the copy of the same strings "on purpose" because I often use poedit which, by default, puts all the empty strings on the top... By doing this I can see the (new) untranslated strings easily ;=)

mintty:
Bien. Look at my other comment and how Super and Hyper are squeezed into the menu:
grafik
That's why I'd suggest to keep them abbreviated.

Gozer-404:
You're right... I didn't get the comment (because I didn't test my changes, sorry).
Lets keep these two abbreviated... but could it be possible to have the keys/checkboxes on two rows instead of one ? or a wider dialog window ?

mintty:
Closing this PR in favour of the new one.
Two lines don't fit. You could make the Options window wider per localization using pseuro-iteim `msgid "100"` (see https://github.com/mintty/mintty/wiki/Tips#adding-translations-to-localization), although that's mainly intended for CJK locales (but also used for Hungarian).

BrianInglis:
Why not just expand the space for menu and dialogue text to accommodate double width characters so there is also more space for localization translations? 
Then you could remove the scaling code and reduce complexity.

mintty:
I think the scaling option makes sense as there are languages that take up more space than others.

Issue #1208, opened 2023-04-02
Selection of words with leading '.' broken since "support nested parentheses" change


clayne:
### Problem:

If one tries to select a word from a line like:

```bash
clayne@a4-sfx:~ $ find -maxdepth 1 -type f -name '.bash*'
./.bashrc
./.bash_history
./.bash_profile
```

The selection will ignore the leading `.` leaving only `/.bashrc` in the selection buffer. This is new (and surprising) behavior. In efforts of figuring out what the problem was, I bisected between `3.0.0` and `HEAD` and found:

```
656d58bc5625eacf3eeb95d541a64c580d703f83 is the first bad commit
commit 656d58bc5625eacf3eeb95d541a64c580d703f83
Author: mintty <mintty@users.noreply.github.com>
Date:   Mon Feb 13 00:00:00 2023 +0100

    hover and open URL: support nested parentheses (#1196)

 src/termmouse.c | 59 +++++++++++++++++++++++++++++++++++++++++++++++++++++----
 1 file changed, 55 insertions(+), 4 deletions(-)
bisect found first bad commit
```

From this version:
```bash
clayne@a4-sfx:~/git/mintty (master=) $ git tag --contains 656d58bc5625eacf3eeb95d541a64c580d703f83
3.6.4
```

### Repro of bisect environment:

The bisect script I used as `/tmp/mintty-bisect`:

```bash
#!/bin/bash

set -e
set -o pipefail

cd ~/git/mintty \
       	&& make clean \
       	&& make -j16 exe \
       	&& bin/cygwin64/mintty -h always bash -c 'echo ./foo.bar' \
       	|| exit 125

echo 'Press enter for GOOD and any other character for BAD' 1>&2
IFS='' read -rn1 tmp && [[ -z "$tmp" ]]
```

And this command to kick it off:
```bash
cd ~/git/mintty \
       	&& git bisect reset \
       	&& git bisect start HEAD 3.0.0 \
       	&& git bisect run /tmp/mintty-bisect
```

Note: some builds will naturally fail depending on how recent the gcc version is on the host (`-Werror`).

BrianInglis:
WJFFM - Works Just Fine For Me:
```
$ mintty -v
mintty 3.6.4 (x86_64-pc-cygwin)
© 2023 Thomas Wolff, Andy Koppe
License GPLv3+: GNU GPL version 3 or later
There is no warranty, to the extent permitted by law.
$ find -maxdepth 1 -name '.bash*'
[...]

PR #1209, opened 2023-04-03
Fix state machine with selection involving URIs

clayne:
* scheme_state_resolve: New function to handle state machine duties when dealing with word selection involving URIs. The previous methodology did not take into account multiple slashes being needed as part of the 'authority' section of the URI suceeding the scheme itself. As such, sequences like "./.bashrc" would stop on the first slash as '.' wasn't considered alpha-numeric and would break the state machine. Also make use an enum for proper state identifiers rather than raw chars.

* sel_spread_word: Make use of scheme_state_resolve.


clayne:
output

src/termmouse.c
```diff
@@ -40,29 +92,10 @@ static char scheme = 0;
...
-      else if (c == ':') {
-        if (scheme == '/')
-          scheme = ':';
-        else
-          scheme = 0;
-      }
-      else if (c == '/') {
-        scheme = '/';
-      }
-      else if (scheme)
```
clayne, line 62: If not wanting to use the "proper" state machine in my PR, the "correct" fix on line 62 would probably be:

```c
else if (scheme == 's')
```
and probably change line 50-51 to:
```c
else
  scheme = 0;
```

That way the selection loop stops only if the state has been elevated to scheme.

mintty:
The change in line 62 without the additional line 50 patch seems to catch more cases of my test list (below) but it's not perfect yet...
And yes, I'd prefer the smaller change...
```
 http://a.b.c/def::ghi
 abc(def)       abc::def

 http://a.b.c/def(ghi)
 (http://a.b.c/defghi)
 http://a.b.c/defghi)
 #http://a.b.c/defghi)
 http://a.b.c/def?ghi=jkl#mno)
 http://user:pw@a.b.c/def?ghi=jkl#mno)
x http://a.b.c/def(ghi)
x (http://a.b.c/defghi)
x http://a.b.c/defghi)
 "http://a.b.c/def ghi"
 http://a.b.c/def::ghi

 http://a.b.c/def(g[h]i)
 {http://a.b.c/defghi}
                                                                 http://a.b.c/defghi

./.bashrc
./foo
http://www.google.com
b/foo
file:///dir/path
/alb
(http:///www.google.com)
http:///www.google.com/maps

**Comment by clayne** [comment]
I see the following issues:

1. `()` has no state management for open vs closed, they're just checked generically with `strchr`.
2. `/` really should have its own state for an opening slash as the spec says `//` must be present.
3. The parsing is done in reverse, making state management that much harder.
4. None of this works for parens in URIs unless the user modifies their own `WordChars`.
5. Even with the `WordChars` change, 2 of the 3 problem cases select reasonably fine for me given the constraints.

Without a real URI parsing SM, preferably in the forward direction, I cannot see how this can be done for parens support without hitting various edge cases. In the meantime it would certainly be preferable if it weren't breaking the normal use cases (e.g. `./.bashrc`).

**Comment by mintty** [comment]
1. () state management is in lines 86 and 92, including nesting support (level). Parenthesis styles are currently not distinguished
2. / is considered in scheme detection, not in the path but a full-blown URL parser isn't really needed, as Brian points out
3. Backwards parsing is due to the use case of clicking inmidst a URL. Could be handled otherwise but I'm not sure that would make it easier (also consider line wrapping...)
4. and 5. I'm focussing on achieving a good user experience without having to configure WordChars.

We have two variants of an implementation now (above) and I'll run a test session when I find the time.

**Comment by mintty** [comment]
The fix shall be as below. If you want, you can submit it as your PR. Wating ~1 day.

```
diff --git a/src/termmouse.c b/src/termmouse.c
index dca3ca18..f7d8e28b 100644
--- a/src/termmouse.c
+++ b/src/termmouse.c
@@ -59,7 +59,7 @@ static char scheme = 0;
       else if (c == '/') {
         scheme = '/';
       }
-      else if (scheme)
+      else if (scheme == 's')  // #1209 / #1208
         break;
       else
         scheme = 0;
diff --git a/wiki/Changelog.md b/wiki/Changelog.md
index 0d901a56..0f205f8b 100644
--- a/wiki/Changelog.md
+++ b/wiki/Changelog.md
@@ -1,3 +1,6 @@
+Terminal features
+  * Fix file link detection (#1208), tweak URL detection (#1209).
+
 ### 3.6.4 (25 Mar 2023) ###

 Terminal features

mintty:
Release 3.6.5.

PR #1219, opened 2023-06-11
Restore ability to open Windows paths

ak2:
The ability to open Windows filenames through Ctrl+click or the Open menu command got lost when removal of backslash quotes for to-be-opened filenames was added in win_open to support opening filenames with escaped spaces and special characters in them.

Restore Windows filename support by only treating backslashes before characters that are special to Unix shells as quotes. (Many of those are illegal in Windows filenames anyway.)

Assume that paths starting with a double backslash are Windows UNC paths.

The ability to open paths with spaces had actually itself got lost because code in term_open truncated the path at the first non-leading space, so remove that.

Also avoid casting away const in win_open, and tweak some comments.


mintty:
Release 3.6.5.

PR #1224, opened 2023-06-25
Restore keyboard to alphabet mode on <Escape>/<Enter> key

qqkookie:
This PR will add new feature to restore keyboard IME state to Latin alphabet mode on  Escape/Enter key.

For non-Latin keyboard user who uses IME to enter native text like CJK script, entering \<Escape> or \<Enter> keypress will restore CJK input method editor (IME) state to Latin alphabet input mode or alphanumeric mode automatically. This spares user of pressing CJK compose key explicitly to enter English  shell command line or alphabet editor key after pressing Enter or Escape key. 

This feature is user-configurable and turned on/off in Options/Keys setting section. This feature does not affect Latin alphabet keyboard like US-English/ASCII or European keyboard. Copy and pasting multi-line CJK-text is not affected.


mintty:
Automatic switching of IME mode to alphanumeric interferes with use cases like fullscreen text editing, see #1375. Now this is a design issue. I guess some tricky way of more clever automatic switching should be found. Since I am not an IME user, do you have any idea perhaps?

PR #1245, opened 2023-12-10
Support lone Alt key presses in KeyFunctions

ak2:
A lone Alt key press in Windows applications normally brings up the application menu bar (which is why its virtual keycode is VK_MENU). There's no such menu bar in mintty, which relies on the context menu and window menu instead, so make lone Alt key presses available for mapping in the KeyFunctions setting.

The action is triggered when the Alt key is released, but only if no other modifier was down when the Alt key was pressed, and if nothing else was pressed in-between.

Also make sure that AltGr does not trigger this action or Alt code processing, unless the AltGrIsAlsoAlt setting is enabled.

One use case for this is to map Alt to the context menu via the 'menu-text' action, to make it easier to reach on keyboards that don't have the context menu key.

Another is to map it to ^[, i.e. the Esc key code, which might be convenient for vim users. It also makes plenty of sense as Alt modifier combinations normally prefix the modified key with ^[.

As the Esc key's default function can be replaced in this way, also make unmodified Esc key presses available for mapping in KeyFunctions. One use case for that is to map it to a non-letter control character such as ^\, and then use that as the interrupt character in the stty settings, instead of ^C. That gives Esc its usual Windows meaning of interrupting or cancelling actions, and makes ^C available for copying.

### DIFF
--- a/docs/mintty.1
+++ b/docs/mintty.1
@@ -2966,9 +2966,9 @@ attached with a "+" separator, e.g. \fBCS+F9\fP;
 or a prefix "*" that applies the definition for all modified variations 
 of the key, e.g. \fB*Menu\fP
 .br
-\(en special keys (some of which do not occur on typical keyboards) 
-and keypad keys, as well as modified special and keypad keys; 
-Space, Enter, Esc and Tab are only considered with at least one modifier
+\(en special keys (some of which do not occur on typical keyboards)
+and keypad keys, as well as modified special and keypad keys.
+Space, Enter and Tab are only considered with at least one modifier.
   Esc, Tab
   Enter, KP_Enter
   Space, Back
@@ -2976,6 +2976,7 @@ Space, Enter, Esc and Tab are only considered with at least one modifier
   ScrollLock, NumLock, CapsLock
   (PrintScreen)
   LWin, RWin
+  Alt (when used on its own)
   Menu
   Insert, KP_Insert
   Delete, KP_Delete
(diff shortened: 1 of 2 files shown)


mintty:
> One use case for this is to map Alt to the context menu via the 'menu-text' action, to make it easier to reach on keyboards that don't have the context menu key.

Added an example for this to the manual.

> As the Esc key's default function can be replaced in this way, also make unmodified Esc key presses available for mapping in KeyFunctions. One use case for that is to map it to a non-letter control character such as ^, and then use that as the interrupt character in the stty settings, instead of ^C.

This can be achieved more directly and decoupled from actual ^C settings by
`KeyFunctions=Esc:break`
Not added as an exmple though.

ak2:
> > One use case for this is to map Alt to the context menu via the 'menu-text' action, to make it easier to reach on keyboards that don't have the context menu key.
> 
> Added an example for this to the manual.

Thanks!

> > As the Esc key's default function can be replaced in this way, also make unmodified Esc key presses available for mapping in KeyFunctions. One use case for that is to map it to a non-letter control character such as ^, and then use that as the interrupt character in the stty settings, instead of ^C.
> 
> This can be achieved more directly and decoupled from actual ^C settings by `KeyFunctions=Esc:break` Not added as an example though.

That doesn't work with ssh though, because the SIGINT is delivered to the foreground process group of the pty that mintty created, thus killing ssh itself, whereas control chars are sent to the remote pty.

mintty:
Thanks for the SIGINT hints, I just had the idea to add a user-definable function `intr` which does exactly what you say `screen` does.

mintty:
Released 3.7.1.

PR #1246, opened 2023-12-10
Fix option index overflow in remember_file_options

ak2:
There are more than 256 config options (!), which means that using a uchar is no longer sufficient for storing the index of options found in the config file or changed in the UI.

The overflow meant that for example a BoldBlue setting in the config file (which currently has option index 262) turned into a TekForegroundColour setting (with index 6) when changing some unrelated setting in the config dialog and saving.


mintty:
Hmm, I wonder what obscure behaviour this may have caused in the past.

PR #1251, opened 2023-12-30
docs: Escape hyphens

### LINKED ISSUE #1250: `-` is shown as U+2010 in the manpage
Recently, `man mintty` shows `-` (U+002D, HYPHEN-MINUS) as `‐` (U+2010, HYPHEN).
So, searching for any options on the manpage (using the `/` command) doesn't work. (E.g. `/--WSL`)

E.g.:
https://github.com/mintty/mintty/blob/91bb8cad25f3856bc46f55d60142768a7cceac3f/docs/mintty.1#L36
`-` in this line should be escaped by `\` as follows:
```
implies a \fB\-\-WSL\fP\fI[=distro]\fP parameter.
```

k-takata:
Close #1250

### DIFF
--- a/docs/mintty.1
+++ b/docs/mintty.1
@@ -8,7 +8,7 @@ mintty \(en Cygwin terminal emulator
 
 .SH SYNOPSIS
 
-\fBmintty\fP [\fIOPTION\fP]... [ \fB-\fP | \fIPROGRAM\fP [\fIARG\fP]... ]
+\fBmintty\fP [\fIOPTION\fP]... [ \fB\-\fP | \fIPROGRAM\fP [\fIARG\fP]... ]
 
 .SH DESCRIPTION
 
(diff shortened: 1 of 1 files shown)


mintty:
I see your patch carefully distinguishes and preserves in-text hyphens. I'm curious: did you check them all manually or is there a suitable tool? (If there is, I'd prefer to keep the unmodified source and add a transformation to the build to prepare the actual manpage.)

k-takata:
Unfortunately, I couldn't find such a tool. So, I did them manually.

Heuristically, if a `-` is sandwiched in the middle of alphabet characters, it is a hyphen. Otherwise, it is an ASCII hyphen-minus. But there are some exceptions.

mintty:
Thank you very much.

BrianInglis:
See [`groff` doc `PROBLEMS`](https://git.savannah.gnu.org/cgit/groff.git/tree/PROBLEMS) for user `man.local` and `mdoc.local` fixes
*NOTE* that changing the default may result in long option names breaking before or within the name, and/or eliminate hyphenation elsewhere. 
Those outcomes may be acceptable if you are content with ragged right margins, and possibly short lines, especially in translations to agglutinative languages such as Deutsch/German or Suomi/Finnish. 
This file and licences are not yet included in Cygwin groff package: I will request they be included.

mintty:
Or the mintty man page could simply add
`.char - \-`
for a general fix. But now there's the distinct modification already, for code-like dashes only. I only hope to keep the discipline to make that distinction.

PR #1257, opened 2024-02-15
Update zh_CN.po

whyXVI:
There's something wrong when displaying the previous translation: the latter part of the text was cut, which causes misunderstandings. Here is a better localization, concise and more understandable.
465a0624e1849de780e3dd8f10b8e5ab

### DIFF
--- a/lang/zh_CN.po
+++ b/lang/zh_CN.po
@@ -637,7 +637,7 @@ msgstr "选中后立即复制(&Y)"
 #. __ Context menu:
 #: config.c:4133 config.c:4166 config.c:4332 wininput.c:684
 msgid "Copy with TABs"
-msgstr "以纯文本复制（包括 &Tab）"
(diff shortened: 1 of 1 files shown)


k-takata:
I think that the space for showing the text "Copy with TABs" should be longer. (Then this PR may not needed.)
The last part of the Japanese translation is not shown either.

mintty:
The space in this area of the Selection pane was still limited to half size as a relic of previous two-column layout. Fixed.
I guess the text change is not required anymore.

mintty:
Other strings further down still don't fit though.
Note that you can expand the width of the Options dialog with a pseudo-message "100".
The Japanese translation for example makes use of this ("121").

mintty:
For those other strings, I have increased the dialog width. To be released soon.

mintty:
Released 3.7.1.

PR #1271, opened 2024-06-20
Some CJK characters don't follow the FontChoice option

k-takata:
The characters listed here (e.g. "「" (U+300C)) don't follow the FontChoice option:
https://github.com/mintty/mintty/blob/78c7d205d586ed87c946b00dc62034ea54c96e7b/src/termout.c#L5224-L5225

Consider the following settings:
```
Font=Anonymous Pro  # ASCII font
FontChoice=CJK:1
Font1=IPAGothic     # Japanese font
```

U+300C is a double width CJK character so `win_char_width()` should return 2 here:
https://github.com/mintty/mintty/blob/78c7d205d586ed87c946b00dc62034ea54c96e7b/src/termout.c#L5222-L5231

However, FONTFAM is not set to `term.curs.attr.attr` at this point, then `win_char_width()` returns 1 and the `TATTR_EXPAND` flag is unintentionally set.  As the result, U+300C is not shown correctly. (The left half of the glyph of U+300C is expanded.)

FONTFAM should be set before calling `win_char_width()`.

### DIFF
--- a/src/termout.c
+++ b/src/termout.c
@@ -4987,6 +4987,10 @@ term_do_write(const char *buf, uint len, bool fix_status)
 
         cattrflags asav = term.curs.attr.attr;
 
+        uchar cf = scriptfont(wc);
+        if (cf && cf <= 10 && !(asav & FONTFAM_MASK))
+            term.curs.attr.attr = asav | ((cattrflags)cf << ATTR_FONTFAM_SHIFT);
+
         switch (cset) {
           when CSET_VT52DRW:  // VT52 "graphics" mode
             if (0x5E <= wc && wc <= 0x7E) {
--- a/src/termout.c
+++ b/src/termout.c
@@ -1248,6 +1248,18 @@ write_ucschar(wchar hwc, wchar wc, int width)
   if (cf && cf <= 10 && !(attr & FONTFAM_MASK))
     term.curs.attr.attr = attr | ((cattrflags)cf << ATTR_FONTFAM_SHIFT);
 
+  // Auto-expanded glyphs
+  if (width == 2
+      // && wcschr(W("〈〉《》「」『』【】〔〕〖〗〘〙〚〛"), wc)
+      && wc >= 0x3008 && wc <= 0x301B
+      && (wc | 1) != 0x3013  // exclude 〒〓 from the range
+      && win_char_width(wc, term.curs.attr.attr) < 2
+      // ensure symmetric handling of matching brackets
+      && win_char_width(wc ^ 1, term.curs.attr.attr) < 2)
+  {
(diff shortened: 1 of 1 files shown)


src/termout.c
```diff
@@ -4987,6 +4987,10 @@ term_do_write(const char *buf, uint len, bool fix_status)
 
         cattrflags asav = term.curs.attr.attr;
 
+        uchar cf = scriptfont(wc);
+        if (cf && cf <= 10 && !(asav & FONTFAM_MASK))
+            term.curs.attr.attr = asav | ((cattrflags)cf << ATTR_FONTFAM_SHIFT);
```
k-takata, line 4992: (If not sure if this is the best place to set FONTFAM, though.)

k-takata:
BTW, should this commented line 
https://github.com/mintty/mintty/blob/78c7d205d586ed87c946b00dc62034ea54c96e7b/src/termout.c#L5224
be as follows (to make it consistent with line 5225)?
```C
            // && wcschr(W("〈〉《》「」『』【】〔〕〖〗〘〙〚〛"), wc)
```
The characters "〒" and "〓" are not a pair.

mintty:
Thanks for the analysis. Before I try to understand it, let's set up a common test case: I could not reproduce this with a Windows preinstalled CJK font - which of them are actually Japanese?
The primary ("ASCII") font does not matter, I assume.
After downloading IPAGothic, the character does not show up at all with unpatched mintty. With the patch, it displays but is not expanded. I guess that's not what you observed.

k-takata:
> I could not reproduce this with a Windows preinstalled CJK font - which of them are actually Japanese?

It can be also reproduced with "MS Gothic".

> the character does not show up at all with unpatched mintty.

Yes, it is the same in my side. Because the left half of the "「" glyph is blank. The blank part is expanded and shown as a double-width blank.
You can also test with "」" (U+300D). The left half of this glyph is not blank, so you might be able to see the issue more clearly.

> With the patch, it displays but is not expanded.

Yes, it is the expected behavior.

mintty:
> (If not sure if this is the best place to set FONTFAM, though.)

scriptfont is called again in write_ucschar so it appears to be a bit redundant (although it does not impose measurable slowdown).
Maybe the whole TATTR_EXPAND check and setting should be moved down into write_ucschar. I'll try that tomorrow.

mintty:
> I'll try that tomorrow.

as attached: 
[1271.patch](https://github.com/user-attachments/files/15922786/1271.patch)

Feel free to convert your PR accordingly.

k-takata:
Thank you. I've confirmed that your patch works fine.

mintty:
Released 3.7.2.

PR #1274, opened 2024-06-27
Update EditorConfig settings

k-takata:
I had to change my editor settings every time when I edit the mintty source code.
This PR solves it.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,3 +2,7 @@ root = true
 
 [*]
 trim_trailing_whitespace = false
+
+[*.[ch]]
+indent_style = space
+indent_size = 2


mintty:
Wouldn't these settings fall back to a global .editorconfig if you have one?
How about simply dropping the line
root = true
? I don't remember why I included it.
Asking for curiosity, merging the PR anyway as it seems reasonable.

k-takata:
Thank you for merging.

The basic idea of EditorConfig is to use the same settings among the members of a project.
So, having a global .editorconfig may not match with the concept. Dropping the "root = true" line may not work as expected.

PR #1292, opened 2024-10-15
Initialise the timestamp earlier

k-takata:
When opening a new tab, it is shown at the left-most tab, and after a while, it is moved to the right-most part since
fa1f16bc772b049872ef2a1ada85dbfd0f1dc898.

Initialise the timestamp earlier so a new tab is shown at the right-most part.

### DIFF
--- a/src/winmain.c
+++ b/src/winmain.c
@@ -7798,6 +7798,9 @@ static int dynfonts = 0;
   win_synctabs(4);
 #endif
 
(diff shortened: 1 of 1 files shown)


mintty:
Thank you, I had not noticed the glitch (not being a tabbar user myself).

mintty:
Released 3.7.7.

PR #1293, opened 2024-10-17
Improve tab reordering

k-takata:
The old implementation didn't correctly detect the end of dragging if
the left mouse button was released out of the tab bar.  Sometimes, tabs
were highlighted unintentionally. Or even crashes occurred.

* Use SetCapture()/ReleaseCapture() to detect the end of dragging.
* Change the mouse cursor while dragging.
* Change the tab order in real-time.

### DIFF
--- a/src/wintab.c
+++ b/src/wintab.c
@@ -233,81 +233,6 @@ static HCURSOR hcursor = NULL;
     ReleaseCapture();
     dragidx = -1;
   }
-#if 0
-  else if (msg == WM_SETCURSOR && dragmsg && LOWORD(lp) == HTCLIENT) {
-    //printf("WM_SETCURSOR lo %02X hi %02X\n", LOWORD(lp), HIWORD(lp));
-    if (HIWORD(lp) == WM_MOUSEMOVE) {
-      // drag during tab reordering; display visual feedback
-
-      // cannot determine current cursor target tab via TCITEM
-      // (like #ifdef determine_tab_index_by_wnd_in_item above)
-      // as the index remains the click-and-drag tab
-
-      // enquire cursor position; derive drop target item index
-      POINT p;
-      if (GetCursorPos(&p) && ScreenToClient(hwnd, &p)) {
-        int x = p.x - xoff;
-        int dropidx = x / curr_tab_width;
-        //printf("%d:%d (pw %d) x %d drop %d\n", (int)p.y, (int)p.x, curr_tab_width, x, dropidx);
-        bool redraw_tab = false;
-        if (dropidx != targidx && targidx >= 0) {
-          // clear hover highlighting of previous drop target
-
-          // clear visual indication via RedrawWindow/WM_DRAWITEM
-          targidx = -1;
-          redraw_tab = true;
-        }
-        if (dropidx != dragidx) {
-          // visual indication of tab dragging: hover highlighting
-          // enquire tab rect to calculate relative cursor position
-          RECT tr;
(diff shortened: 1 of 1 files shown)


mintty:
Thank you very much for handling issues that I hadn't noticed, and improving more.
I'm however missing somehow the dynamic colour background feedback in the tabbar which I liked (part of code that you removed).

mintty:
Not sure whether we're talking about the same thing, and how vim would be involved.
The current tab is always highlighted, yes. When click-and-dragging, in my previous implementation, while dragging over other tabs, their tabbar colour would change from normal to dark blue gradually, with a peak in the middle.

k-takata:
> and how vim would be involved.

This is because I was referring to Vim's code when I implemented this PR:
https://github.com/vim/vim/blob/6c2fc377bfbfb6f1a46b1061413cd21116b596ed/src/gui_w32.c#L8420
(partly implemented by me)



> When click-and-dragging, in my previous implementation, while dragging over other tabs, their tabbar colour would change from normal to dark blue gradually, with a peak in the middle.

In my implementation, tabs are reordered soon if the mouse cursor goes over another tab. So, there might be no chance to set the background color. If we set a threshold for it, setting the color might be possible.
(The above Vim implementation has a threshold, though.)

mintty:
I have reactived dynamic tab highlighting. First I fixed the crash condition, then I adapted the approach to the new paradigm of dropping tabs right away while dragging. The dynamic colouring is a substitute of dynamic visual tab placement as other applications (browsers) do. Colour could be chosen to something different, see comments after line 360.
I'd appreciate if you check this.

k-takata:
I've tried your latest commit. It works fine. Thank you for revising my PR.

src/wintab.c
```diff
@@ -189,88 +191,48 @@ static int targpro = -1;
...
-          targidx = dropidx;
-          redraw_tab = true;
-        }
-        if (redraw_tab)
-          RedrawWindow(tab_wnd, 0, 0, RDW_INVALIDATE | RDW_UPDATENOW);
-      }
-    }
-    else if (HIWORD(lp) == WM_LBUTTONUP) {
-      // drop tab while dragging for tab reordering
+        if (abs(p.x - xpos) > GetSystemMetrics(SM_CXDRAG)) {
```
k-takata, line 212: For better high DPI support, it might be better to use `pGetSystemMetricsForDpi()` instead of `GetSystemMetrics()`.

mintty: Sounds proper in theory but it doesn't seem to make a difference. Checking my previous use of the Dpi version I did some tests with it but it's currently not in use anywhere in mintty.

k-takata: > Sounds proper in theory but it doesn't seem to make a difference.

`GetSystemMetrics(SM_CXDRAG)` normally returns a small value (e.g. 3). Even if the value is multiplied by the DPI ratio, it will be still small. So, it might be difficult to notice the difference.

mintty:
I added the value to the debug output (winmain.c line 3071) and values were the same for me (switching between 125% and 175% zoom factor).

k-takata:
Oh, you are correct. I didn't know that some values are unaffected by the DPI ratio.

mintty:
I've further enhanced tab dragging by swiping the tab button. Comments welcome.
I couldn't swipe the border though - it seems to be drawn implicitly.

mintty:
I also changed the tab colour while dragging. Planning a release this weekend.

mintty:
Released 3.7.7.

mintty:
Released 3.8.0. Fixed colour glitches in the visual feedback during tab reordering.

PR #1294, opened 2024-10-21
Restore focus after tab reordering

k-takata:
Set the focus to the main window after tab reordering is finished.

Users couldn't type texts or commands after tab reordering without clicking the main window.


mintty:
Oops, I hadn't noticed. Thank you.

Issue #1303, opened 2024-12-14
sync theme with Windows Theme and Darkmode


Klauswk:
Hello,

Inside mintty, it would be nice to have an option allowing sync of the theme with Windows 10/11 if the default theme is set to none.

For example:
![image](https://github.com/user-attachments/assets/6ad20e9a-e6a5-4c2e-8e49-132023679e10)

Then, when changing the theme in Windows settings, it automatically changes inside mintty as well.

We could also allow the user to predefine the dark/light theme:

![image](https://github.com/user-attachments/assets/5dda2fed-a5c5-4d4c-9db4-cbf1c6a605af)

What do you guys think?

mintty:
Did you try setting `UseSystemColours=true`? Is that what you want? If not, which further properties would you like to be taken from Windows?
And if so, how would they be retrieved and how would Windows light/dark mode choice be detected?

BrianInglis:
Check hourly or when settings change?
```
$ regtool list -v /proc/registry/HKEY_CURRENT_USER/Software/Microsoft/Windows/CurrentVersion/Themes/Personalize/
ColorPrevalence (REG_DWORD) = 0x00000000 (0)
EnableTransparency (REG_DWORD) = 0x00000000 (0)
AppsUseLightTheme (REG_DWORD) = 0x00000001 (1)
SystemUsesLightTheme (REG_DWORD) = 0x00000001 (1)
```
Some APIs may not yet be available in *w32api-headers*. 
Checking for dark/light mode by checking *foreground* colour seems non-intuitive:
[Support Dark and Light themes in Win32 apps](https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/ui/apply-windows-themes)

Remember that *high latitudes* get dark earlier in winter and later in summer; those seasons depend on *sign of latitude*; **18:00** local time may be a reasonable general default for switching, unless you have time zone DST dates, check the sign of change, and bump the switch time by a proportional delta changing sign in the middle of the DST period, perhaps corrected by or proportional to the *Equation of Time*, depending on ingenuity?

Klauswk:
> Did you try setting `UseSystemColours=true`? Is that what you want? If not, which further properties would you like to be taken from Windows? And if so, how would they be retrieved and how would Windows light/dark mode choice be detected?

[...]

PR #1305, opened 2024-12-24
Add support to sync Windows Theme with mintty

Klauswk:
As discussed in issue #1303, this PR allows Mintty to sync when the user changes the "app mode" in Windows 10/11.

1. A single new option was added, "Dark theme"
2. The size of the dropdown "Theme" was changed to match the size of the new dropdown
3. Add a new option in the config called "dark_theme", with the default value as empty.

image

A few points on my side:

1. There might be a better place to add the check for the value change, I added it to the src/winmain.c for convenience.
2. theme_handler wasn't ready to accept more than one dropdown config, I did my best to change the least amount of code as possible, but my test might not have covered all the scenarios
3. One config being called "theme_file" and the other "dark_theme" might not be the best, suggestions?

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -1881,8 +1881,8 @@ apply_config(bool save)
       if (changed)
         remember_file_option("apply", i);
     }
-
   }
+
   copy_config("apply", &file_cfg, &new_cfg);
   if (wcscmp(new_cfg.lang, cfg.lang) != 0
       || (wcscmp(new_cfg.lang, W("=")) == 0 && new_cfg.locale != cfg.locale)
--- a/src/config.c
+++ b/src/config.c
@@ -64,7 +64,7 @@ const config default_cfg = {
   .theme_file = W(""),
   .background = W(""),
   .colour_scheme = "",
-  .use_windows_theme = false,
+  .dark_theme = W(""),
   .transparency = 0,
   .blurred = false,
   .opaque_when_focused = false,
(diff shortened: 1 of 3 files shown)


mintty:
Thanks for working this out. Initial comments:

I’d keep the position of new attributes in sync in config.h and config.c initialisation, and place dark_theme after theme_file.

Function name set_new_cfg is quite generic, handling only themes.
The handling of either theme attribute could also be supported by two wrapper functions. This would also spare the comparison of label strings, which isn’t done elsewhere I think. Also strcmp with "&Theme" will fail if the UI is localised.
I guess I’ll check details next year...

Initialisation in winmain:
If theme_file is configured, dark mode is ignored.
If dark_theme is configured, its value is ignored.
I guess both is not intended.
Also not sure why you set two specific themes as respective default.
And instead of setting „kohlrausch“, theme should be cleared to the default.

Dynamic adaptation in winmain: I think this shouldn’t be checked continuously but hooked to a suitable Windows message like WM_WININICHANGE.

Style: Please don’t remove empty lines where useful for structuring and keep some space after if, while, etc.

Klauswk:
Thank you for the review.

1. Fixed theme config order
2. Added missing space after if/while
3. Rollback the new line removal

mintty:
Hi, sorry for the delay.
Did you/Do you intend to continue work on this and provide another PR?
Otherwise, as I cannot accept selected parts of a PR, I would offer this:
I continue to work this out according to my own review comments, then commit it with your authorship, or, if you prefer automatic credits for the contribution, offer the result back to you to submit as a PR.

Klauswk:
Sorry for the delay as well.

You can go ahead and finish this PR, since some parts might require more knowledge of the system, and right now, unfortunately, I can't put in the time.

As for the strategy, using the authorship seems more reasonable, since both of us would have worked on it.

Thank you for your time.

mintty:
I have now integrated the dark mode theme with the following changes:
* renamed DarkTheme → ThemeDark, to align better with ThemeFile
* fixed logic to choose darkmode theme in apply_config and main
* dropped continuous checking in main loop
* changing theme instead when Windows dark mode changes
* dropped config in Options dialog for now as some details are still to be considered
To be checked: interworking with colour scheme configuration (e.g. use of Color Scheme Designer)

mintty:
Released 3.7.8 with option ThemeDark, not yet in interactive Options dialog.

PR #1328, opened 2025-07-26
Simplify and standardize cygport usage a bit

jon-turney:
* Use cygport to construct source and binary package archives
* Use cygport to upload them
* Align mintty.cygport with handwritten hint files, which can be removed


jon-turney:
Notes on future work: Maybe sed VERSION and RELEASE into mintty.cygport, rather than passing those via the filename it's renamed to?

mintty:
Thanks for this endeavour and particularly for integrating it with the existing Makefile so I can keep my workflow :)

PR #1339, opened 2025-10-18
manifest support for per-monitor high DPI support

clin1234:


mintty:
Per-monitor support was added to the code 10 years ago and seems to be working.
What benefit would this addition to the "manifest" have?

mintty:
Yet another question:
Another manifest (https://gist.github.com/emoacht/7e5a026080aeb7eb1b9316f5fe7628da) has extended values for both parameters. Does that mean anything?

clin1234:
> Yet another question: Another manifest (https://gist.github.com/emoacht/7e5a026080aeb7eb1b9316f5fe7628da) has extended values for both parameters. Does that mean anything?

Those are fallback settings for older versions of Windows in case they don't support the newer high DPI settings

mintty:
Thanks so far. The most delicate question I have in mind is the test case to move a window between two monitors with different DPI setting. This should be tested by someone before committing this change which otherwise looks fine.

mintty:
I've tested this in Virtualbox with 2 virtual monitors. Merging, thank you.

PR #1342, opened 2025-11-13
Update fr.po

Gozer-404:


mintty:
Merci! As you localized Boolean values like "yes" etc, please note the warning about this in the wiki Tips page and make sure you tested whether everything works as desired.

PR #1343, opened 2025-11-14
fr.po updated

Gozer-404:
Misc typos + removed "localization optional" when booleans were involved


Gozer-404:
Hi,

I'm not so familiar with GitHub, thus maybe the error was on my side.

Anyway the latest version of the fr.po file has only a few simple changes,
that can probably be merged manually and easily (
https://github.com/Gozer-404/mintty/tree/master/lang)

Regards

Issue #1113, opened 2021-07-31
mintty is slow to grab focus on startup, which interacts badly with focus-stealing prevention


benrg:
There's a human-perceptible delay between the mintty process being created and the appearance of the terminal window on the desktop. It's only a fraction of a second, but it creates a bigger problem. I have a hotkey that starts a mintty terminal, and I tend to press the hotkey and then immediately start typing what I want to go into the terminal. Sometimes, the first keystroke precedes the appearance of the terminal window, so it goes to another window. Then (I assume this is what is happening) Windows' focus-stealing prevention algorithm ironically decides that I really wanted to be typing into that other window, and never gives focus to the new terminal. By the time I notice, a number of additional keystrokes have gone to a window that, in many cases, I can't even see any more because the new terminal has appeared on top of it. Cleaning up after this is a pain.

I'm able to consistently reproduce this by pressing my mintty-launching hotkey and then quickly pressing a random key. If I make the same hotkey launch notepad.exe instead, I can't reproduce it, so I think the problem is mintty being slow to grab focus.

I don't know if fixing this is feasible, but it would be nice.

mintty:
There is hardly any terminal that would start "immediately". There are typically a number of parameters, e.g. command-line parameters, that impact the appearance of the terminal, unlike simple system programs like notepad, which may explain some tiny difference.
Most people would simply wait until a window appears before they type "into it", so this is not a priority issue either.
However it's a challenge and I may find some time to look into it but don't expect miracles.

mintty:
On my system, it takes ~130ms from process start until display of the window.
~80ms are taken by fork(), a notoriously slow system call in cygwin, unfortunately.
Now it might be feasible to create the child process *after* the window is shown, with unforeseen side effects however.

mintty:
Applied a few tweaks to get the window to display earlier:
Font initialisation could be optimised, saving ~40ms.
Drag-and-drop setup can be done after showing the window, postponing another 7ms.
Showing the tabbar (if configured) can also be moved after window display, for another 12ms deferred.
ShowWindow() itself takes 18ms, CreateWindow() 4ms, both unavoidable.
The major delay remains starting the shell with fork() taking 80ms which are out of control of mintty.
I tried creating the shell after displaying the window but it has unpleasant side effects (as commented in the code).

mintty:
Actually it seems to be possible to grab the focus even before showing the window, as a workaround to evade the child creation delay for keyboard handling.
Please try the uploaded repository version.

[...]

PR #1362, opened 2026-04-23
winmain: skip startup SetFocus when window is hidden (-w hide)

zvirja:
I am currently trying to have an experimental plugin for GitExtensions GUI to embed mintty into the app. I am running the mintty hidden using (-w hide), but yet it still steals focus from the GitExtensions and that leads to flickering.

In this PR I am avoiding calling `SetFocus()` if window is hidden. The three startup SetFocus(wnd) calls exist as a workaround for #1113 (keystrokes lost during the focus-acquisition delay on visible startup). But if the window is not visible, that is not the issue.

This PR was AI assisted with Claude Opus 4.7.

### DIFF
--- a/src/winmain.c
+++ b/src/winmain.c
@@ -8419,11 +8419,16 @@ static int dynfonts = 0;
   }
 
   // Grab the focus into the window.
-  /* Do this before even showing the window in order to evade the 
+  /* Do this before even showing the window in order to evade the
      focus delay enforced by child_create() (#1113).
      (This makes the comment below obsolete but let's keep it just in case.)
+     Skip when the window is hidden (-w hide): there is nothing for the
+     user to type into, and focusing a hidden top-level window on a newly
+     started process triggers spurious foreground activation in embedding
+     hosts.
   */
-  SetFocus(wnd);
+  if (cfg.window)
+    SetFocus(wnd);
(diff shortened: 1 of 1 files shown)


mintty:
Makes sense, thank you.

zvirja:
@mintty Thanks for the fast response!

Do you know how quickly it's going to be publish with Git for Windows? Do they usually pick it up fast, or is the cycle really slow?

mintty:
I think they pick up fast, via MSYS2 though which is the project that integrates cygwin stuff and Git is based on.
Only after the mintty release, of course, which I do not plan for quite the next week...

mintty:
You could compile mintty yourself for your purpose, e.g. in MSYS2 and then use it in Git.

zvirja:
I don't want to distribute my own copy - it's easier to wait and use what is bundled with Git. This change is not a critical deal and it looks like it will be there in a month or so, so that's totally fine!

zvirja:
@mintty Do you plan release any time soon? We recently found that we actually need this patch for the app to work right. So I am curious if you maybe release a new version somewhere soon 😊

mintty:
Yes, one open issue in code and a few documentation issues left.

PR #1377, opened 2026-08-26
`isnanf()` fix for 32-bit

vmunix486:
`isnanf()` is not available on 32-bit platforms. I found this out because I tried to compile it on my laptop, and it wouldn't compile.

This is a simple fix. I'm not really sure what the preprocessor directive is for 32-bit, but `i386` worked for me.

While reading what `h` was, it seems to be that it's not defined as a float (there is a line defining it as such, but it is commented out, along with many other single character variables), so I'm pretty sure it can use `isnan()` normally, but I don't know a lot of C, so take that with a grain of salt. It's up to you to do that.

### DIFF
--- a/src/regis.c
+++ b/src/regis.c
@@ -797,7 +797,12 @@ rgbToHsl(float r, float g, float b)
   if (maxcolor == 2)
     h = 4 + (r - g) / (max - min);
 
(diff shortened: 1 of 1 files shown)


vmunix486:
Oops, I just read the README file and learned how to contribute, sorry.

